PR #2, opened 2016-03-04
Reformatting and added docstrings.

ekse:


rhelmot:
The docstring at line 114 looks a little weird, is that what you meant to put?

ekse:
Yep, fixed now.

ekse:
Cool, thanks :)

PR #4, opened 2016-06-21
Added the prologue "mov edi, edi; push ebp; mov ebp, esp"

quoscient:
This prefix is common in Windows binaries.
The set function_prologs was converted to a list to keep the order (important since one prologue is a prefix of another).


ltfish:
Sorry for the long delay. I was really busy last week.

This PR looks good to me. Thanks!

PR #6, opened 2017-02-22
Additional prologues for PPC64 from clang output.

mattrepl:
Added additional cases from clang output. In one case, `mflr r0` is the start of the prologue. In another, arguments from registers (`r3` being the first arg) are immediately stored on the stack.


rhelmot:
Awesome, thanks!

PR #10, opened 2017-04-26
Argument registers and positions for AMD64

mattrepl:
Added argument register position info. Removed registers from argument_registers that are not used for passing arguments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: xenial
+dist: trusty
 env:
   - PY=e ANGR_REPO=simuvex
   - PY=e ANGR_REPO=angr


ltfish:
I'm not against this PR. But a note to angr team: in the long run, we should encode all those information in `SimCC`.

mattrepl:
Yeah, agreed on needing a better place for calling convention info. I needed argument register order for the hack-y variable recovery I've been working on. Speaking of, will take a look at what you're doing.

zardus:
I think I screwed something up with git. I was trying to rebase this on master and now everything has exploded.

zardus:
Ok. I think we have to abandon this PR and migrate to #11.

PR #14, opened 2017-08-13
Add PE relocations

sraboy:
This will be needed for an upcoming refactor in CLE for the PE backend.


ltfish:
Fine by me. I assigned it rhelmot in case he has something to say.

rhelmot:
:+1:

PR #15, opened 2017-08-17
Add Endness constants for little and big endian endness

bennofs:
This way, you will at least get a runtime error about a missing attribute
if you mistype the name (previously, if you wrote IEnd_LE it would be silently
ignored)


rhelmot:
Cool! You can replace all the Iend_* values in archinfo itself using this, if you like.

PR #17, opened 2017-09-28
Expanded Reality support

subwire:
Imagination is boring, let's blow angr's mind and expand the reality of what it can handle.

- Add variable byte-width support
- Add some helpers for converting things
- Support the Third Endness

This shouldn't(tm) break anything, and should probably be merged first.

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -85,9 +85,11 @@ class Arch(object):
     :ivar list lib_paths: A listing of common locations where shared libraries for this architecture may be found
     :ivar str got_section_name: The name of the GOT section in ELFs
     :ivar str ld_linux_name: The name of the linux dynamic loader program
-    :ivar int byte_width: the number of bits in a byte.
+    :cvar int byte_width: the number of bits in a byte.
     :ivar TLSArchInfo elf_tls: A description of how thread-local storage works
     """
+    byte_width = 8
+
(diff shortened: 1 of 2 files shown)


archinfo/arch.py
```diff
@@ -127,6 +130,8 @@ def __init__(self, endness):
                 if hasattr(self.uc_const, reg_name):
                     self.uc_regs[r] = getattr(self.uc_const, reg_name)
 
+        self.byte_width = 8
```
rhelmot: This needs to be a class variable, not an instance variable.

archinfo/arch.py
```diff
@@ -272,6 +279,12 @@ def translate_register_name(self, offset, size=None):
         except KeyError:
             return str(offset)
 
+    def get_register_offset(self, name):
+        try:
+            return self.registers[name][0]
+        except:
+            raise ValueError("Register %s does not exist!" % name)
+
```
rhelmot: What was the purpose of adding this? It doesn't seem to add any value to just saying `arch.registers[name]`, and generally you do not want to separate the size from the offset.

archinfo/arch.py
rhelmot: Also change the `:ivar byte_width` to `:cvar byte_width:`

PR #20, opened 2017-10-17
Suggestion: Add fptag0-fptag7 index for FPU tag word（fptag）

YHZX2013:
As code`pmullw mm0, mm3`:
```
01 | PUT(ftop) = 0x00000000
02 | PutI(136:I8x8)[0x00000000,0] = 0x01
03 | PutI(136:I8x8)[0x00000000,1] = 0x01
…………
09 | PutI(136:I8x8)[0x00000000,7] = 0x01
…………
```
When the breakpoint (regwrite/regread) is triggered, the index of register is between 136-143.....

Also reference to the struct in libvex:
```
/* FPU */
ULong guest_FPREG[8];    /* 72 */
UChar guest_FPTAG[8];   /* 136 */
```
Thanks.


rhelmot:
The reason we don't have the individual registers labeled except as mmxX is because what you usually think of as "number 0" according to common labeling schemes actually moves around in the register file depending on the value of `ftop`. If you're okay with the semantics of these registers being that they refer to "the offset in the register file" and not to "the tag corresponding to the nth fpu stack element", then I will merge this.

YHZX2013:
Yes, this really could lead to misunderstanding of the real situation of fptags as intel-manual says:
```
Each tag in the x87 FPU tag word corresponds to a physical register (numbers 0 through 7). The current top-of- stack (TOP) pointer stored in the x87 FPU status word can be used to associate tags with registers relative to ST(0).
```
Suggestion for spliting the fptag reg only considering for convenient using……

By the way, it seems that VEX is not good for x87 support? Such as 80bit double extended-precision floating-point supporting (regs, IRType and IROP……).

Thanks for reply.

rhelmot:
vex is indeed poor for x87 support. This isn't really a limitation of the IR itself, just valgrind's lifter is missing support for it. I think what happens is that it treats all 80 bit floats as just doubles, and then has a special dirty op to convert to/from the 80 bit format when it's stored in memory.

Another issue you will run into is that z3 doesn't have support for floating point theory of the long double (80bit) floats.

YHZX2013:
Okay, marked.
Fortunately, x87 program seems being less and less, and gradually replaced by the sse program......

zardus:
I think Z3 does support configurable mantissa and exponent bitwidths for its floating points (http://z3prover.github.io/api/html/namespacez3py.html#a567a5422d0233937d12b4280177cc153), but a) for all I know, 80-bit floats are somehow different (@jmgrosen and @pcgrosen are the true masters for that, probably) and b) VEX's handling of them via conversion in and out of doubles makes that pretty irrelevant anyways.

pcgrosen:
80-bit x87 floats are a tad bit different; they have an explicit "normalized" bit (bit 63) instead of the so-called "hidden" bits of single and double formats. As a result, the significand is essentially one bit smaller than what you might expect, so you may have to model this in Z3 by using 63 (instead of 64) for the `sbits` parameter.

YHZX2013:
Thanks all. 
@rhelmot Close this pull request?

PR #21, opened 2017-10-31
Add some arm registers aliases used in capstone

YHZX2013:
As title, when using function cs_reg_name()...


zardus:
So we have a problem here... `ip`, in angr parlance, is `instruction pointer`, and `state.regs.ip` is accessed in places all over the framework. We can't have an `ip` mean something else (in this case, apparently, r12 is ip as in the intra-procedure call scratch register).

YHZX2013:
@zardus Yes, this line has been deleted, my fault.

zardus:
Awesome, thanks!

PR #24, opened 2017-12-08
Add Soot support.

ltfish:


archinfo/arch.py
```diff
@@ -91,11 +93,15 @@ class Arch(object):
     byte_width = 8
 
     def __init__(self, endness):
-        if endness not in (Endness.LE, Endness.BE, Endness.ME):
-            raise ArchError('Must pass a valid VEX endness: Endness.LE or Endness.BE')
+        if self.vex_support:
+            if endness not in (Endness.LE, Endness.BE, Endness.ME):
+                raise ArchError('Must pass a valid VEX endness: Endness.LE or Endness.BE')
```
rhelmot: endness is supposed to be independent from VEX, that's why we have the enum... Can you make this a better condition? has_memory?

archinfo/arch.py
```diff
@@ -305,6 +311,13 @@ def library_search_path(self, pedantic=False):
             path = sum([[x + 'tls/${ARCH}/', x + 'tls/', x + '${ARCH}/', x] for x in path], [])
         return map(subfunc, path)
 
+    vex_support = True
+    unicorn_support = True
```
rhelmot: both of these can be removed and inferred through whether vex_name and qemu_name are None

archinfo/arch.py
```diff
@@ -305,6 +311,13 @@ def library_search_path(self, pedantic=False):
...
+    address_types = (int, long)
+    function_address_types = (int, long)
+    supported_register_types = { 'SimSymbolicMemory', 'SimFastMemory' }
+    supported_memory_types = { 'SimSymbolicMemory', 'SimFastMemory', 'SimAbstractMemory' }
```
rhelmot: We can't name angr objects from archinfo...

archinfo/arch.py
rhelmot: has_memory can be inferred from memory_endness is None, like the below?

archinfo/arch.py
ltfish: Yeah... Otherwise this information will have to be hardcoded somewhere in `angr`, which is what I wanted to avoid. Let me come up with a better solution.

PR #26, opened 2018-02-04
remove AVR archinfo (now part of angr_platforms now)

bennofs:
This archinfo module conflicts with the one in angr platforms, which makes working on angr 
platforms hard (since the arch module in platforms is overriden by this one).

I think it makes sense to remove this module from archinfo since angr itself also does not support
AVR, or is this necessary for some reason?


zardus:
That works for me. @subwire, any objections? Also, I think we should open a WIP PR for the AVR and SH4 (cc @pwnslinger) branches in angr-platforms, for more visibility.

ltfish:
Fine by me, too. @subwire ?

ltfish:
I'm going to go ahead and merge this in now. Thank you!

PR #27, opened 2018-03-09
Keystone integration

tyb0807:
Part of angr/angr#892
@rhelmot

### DIFF
--- a/archinfo/arch_ppc32.py
+++ b/archinfo/arch_ppc32.py
@@ -1,12 +1,24 @@
+import logging
+
+l = logging.getLogger("archinfo.arch_ppc32")
+
 try:
     import capstone as _capstone
 except ImportError:
+    l.warning("Unable to import Capstone...")
     _capstone = None
 
+try:
+    import keystone as _keystone
+except ImportError:
+    l.warning("Unable to import Keystone...")
+    _keystone = None
+
 #try:
-#   import unicorn as _unicorn
+#    import unicorn as _unicorn
 #except ImportError:
-#   _unicorn = None
+#    l.warning("Unable to import Unicorn...")
+#    _unicorn = None
 
 from .arch import Arch, register_arch, Endness
 from .tls import TLSArchInfo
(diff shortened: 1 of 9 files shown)


ltfish:
Please modify `setup.py` and `requirements.txt` to add a dependency to keystone.

rhelmot:
Disagree? It's an optional dependency

ltfish:
> Also, it's already checking if _keystone is truthy before using it...

`def asm()`? I don't think there is a check there.

rhelmot:
https://github.com/angr/archinfo/pull/27/files#diff-d6b6db071fabd3fe806f4451dca2af0cR127

_ks_arch will be None if keystone is not installed

ltfish:
I would explicitly check `_keystone` instead of relying on `_ks_arch` there, for a better readability. This is definitely less important.

rhelmot:
mhm. there should be a better error message if keystone is not installed.

ltfish:
How about: When `asm()` is called and keystone is not installed, print out something meaningful?

rhelmot:
That's exactly what I was trying to say!

tyb0807:
@ltfish @rhelmot is it better now?

archinfo/arch_ppc32.py
```diff
@@ -1,12 +1,21 @@
...
     _capstone = None
 
+try:
+    import keystone as _keystone
+except ImportError:
+    _keystone = None
+
 #try:
-#   import unicorn as _unicorn
+#    import unicorn as _unicorn
```
ltfish: This change is not necessary.

archinfo/arch_ppc32.py
```diff
@@ -1,12 +1,21 @@
...
 #except ImportError:
-#   _unicorn = None
+#    _unicorn = None
```
ltfish: Same as above - this change is not necessary.

ltfish:
Thank you! Other than two small places that you added extra spaces, everything looks good to me!

@rhelmot Comments?

archinfo/arch_ppc32.py
tyb0807: Bizarre... I don't see those extra spaces at all?

ltfish: That line should not be changed. Are you saying you don't see those extra spaces in your local editor? Maybe there is some special setting with your editor?

tyb0807: I know why, the extra spaces are correct since we have '#' at the start of these lines (I assume the correct tab width is 4)

PR #32, opened 2018-03-23
Py3k compatible

tyb0807:
@rhelmot @zardus please take a look at this

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -1,3 +1,5 @@
+from past.builtins import long
+from future.utils import iteritems
 import logging
 import struct as _struct
 import platform as _platform
@@ -206,7 +208,7 @@ class Arch(object):
     def get_default_reg_value(self, register):
         if register == 'sp':
             # Convert it to the corresponding register name
-            registers = [r for r, v in self.registers.items() if v[0] == self.sp_offset]
+            registers = [r for r, v in iteritems(self.registers) if v[0] == self.sp_offset]
             if len(registers) > 0:
                 register = registers[0]
             else:
@@ -362,7 +364,7 @@ class Arch(object):
 
         if pedantic:
             path = sum([[x + 'tls/${ARCH}/', x + 'tls/', x + '${ARCH}/', x] for x in path], [])
-        return map(subfunc, path)
+        return list(map(subfunc, path))
 
     @property
     def vex_support(self):
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -308,6 +308,9 @@ class Arch(object):
             bytelist, _ = self._ks.asm(string, addr)
             if as_bytes:
                 encoding = ''.join(chr(c) for c in bytelist)
+                if not isinstance(encoding, bytes):
+                    l.warning("Cheap hack to create bytestring from Keystone!")
+                    encoding = encoding.encode()
             else:
                 encoding = bytelist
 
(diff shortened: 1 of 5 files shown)


tyb0807:
I don't know why but `test_keystone` does pass on my local machine? Is it because of keystone on pip or something? Because I have keystone installed from source

archinfo/arch.py
```diff
@@ -1,3 +1,5 @@
+from builtins import map, chr, object
```
rhelmot: Why is this necessary?

rhelmot:
The keystone is probably failing because of my hacked solution to add the `as_bytes` and your having fucked with `chr`.

archinfo/arch.py
tyb0807: Well, that's what `futurize` recommend to add to as in here http://python-future.org/imports.html. Let me remove it and retry in Python 3

tyb0807:
Tests pass now! I removed all the `builtins` imports.

rhelmot:
1. I am concerned that you are not applying the appropriate care to these pull requests - you should know by now that `from __future__ import ...` will fail if it's not the first line in the file, and that's currently the case in offsets.py. There might be more stuff like this.
2. I am also concerned about merging this while it can't be really tested under py3k. For example, I know for sure we'll run into issues because of the `chr` stuff on python 3 - it'll return a unicode character, and then we'll be returning unicode strings from `.asm()`.

tyb0807:
1. I thought I had fixed that `import`, but it was only in `monkeyhex` 😞. Really sorry for this....
2. I'm not sure why returning unicode strings from `.asm()` would be a problem. The `as_bytes` parameter is supposed to transform a `list` to a `str`. And also in `angr/block.py` and `angr/engines/vex/engine.py`, `insn_bytes` being a `str` is fine too if I understand it correctly. I copied that part of code from angr and try to run in Python 3 (since we have Pyvex py3k compatible now) and it is ok. But I'm not super sure if it is really safe though.

rhelmot:
I mean... there's a real fundamental difference between "a sequence of bytes" and "a sequence of unicode codepoints". pretty much everything we do in angr is with a sequence of bytes. It doesn't make sense to be passing around machine bytecode as a unicode string. The thing is that python 3 has changed the `str` class from being a sequence of bytes to being a sequence of unicode codepoints - the `bytes` class is meant to do that now. In python 2 `bytes` is a synonym for `str` in order to ease the transition.

tyb0807:
Yes, I know what you mean. I agree that semantically, a `str` (in Python 3 way, so a sequence of unicode code points) should not represent machine byte code. But what I wanted to say is, the only usage of `insn_bytes` (that should be of type `bytes`) is to create an IRSB in pyvex, and the thing is that pyvex does accept it as a `str` (again, in Python 3 way). So the question is, do we still want to change this, I mean, make sure that `insn_bytes` is of type `bytes` in Python 3, and `str` in Python 2?

rhelmot:
I think pyvex accepting a unicode string is asking for disaster:

```python
In [4]: a = ''.join(chr(c) for c in [0x41, 0x42, 0xff])                                                  

In [5]: a                 
Out[5]: 'ABÿ'             

In [6]: type(a)           
Out[6]: str               

In [7]: a.encode()        
Out[7]: b'AB\xc3\xbf'
```

(technically, encode("latin-1") solves this but I desperately do not want to do that without raising an ugly warning at the user)

tyb0807:
Can someone rerun Travis to see if tests pass please? Unfortunately I don't have rights to do this

zardus:
looks like they passed -)

PR #33, opened 2018-04-09
archinfo: python3 compatibility

Mic92:
There could be more, but then something like the `future` package is needed to keep it maintainable.

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -2,6 +2,7 @@ import logging
 import struct as _struct
 import platform as _platform
 import re
+import sys
 from archinfo.archerror import ArchError
 
 l = logging.getLogger("archinfo.arch")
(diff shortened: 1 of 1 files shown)


rhelmot:
Thanks!

/cc @tyb0807

archinfo/arch.py
```diff
@@ -532,7 +538,7 @@ def register_arch(regexes, bits, endness, my_arch):
             raise ValueError('Invalid Regular Expression %s' % rx)
     #if not isinstance(my_arch,Arch):
     #    raise TypeError("Arch must be a subclass of archinfo.Arch")
-    if not isinstance(bits, int):
+    if not isinstance(bits, integer_types):
```
Mic92: mhm. I hope I did not introduce unwanted behavior here.

rhelmot: Shouldn't be. I'd call it a bugfix if anything.

tyb0807:
@rhelmot Thanks for the ping. I still think that `future` package should be the most pythonic way to write Python2/3 code, even when its internal code contains a bunch of conditions as in this PR. And sorry for the late reply.

Mic92:
If I have bigger refactoring to do in this module, I will take this into account.

PR #35, opened 2018-06-10
Add Register class

tyb0807:
Hello,

@rhelmot please take a look at this. I do not know how to make this work with metaclass yet, but please tell me whether I am on the right track.

Thanks for your time


rhelmot:
This looks extremely good. I'm inclined to merge it right now (as soon as the test passes of course) and work on optimizations like the metaclass stuff later.

angrop is the purvey of @salls. Those files serve as a ground truth and get unpickled and compared against the live version, so we could probably cross our fingers that everything was fine and just repickle it. We could also just find-replace the register name, since it's the old ascii pickle format...

PR #41, opened 2018-07-03
Improved Java support.

ltfish:


archinfo/arch_soot.py
```diff
@@ -1,12 +1,17 @@
 
 from .arch import Arch, register_arch, Endness
+import re
+
+import logging
+l = logging.getLogger('archinfo.arch_soot')
```
ltfish: The import sequence:

1. Python built-in packages.
1. Third-party libraries.
1. Other angr libraries.
1. Modules from the current module (archinfo).

Make sure to leave a blank line between each group of imports.

So ideally it should look like the following:

```
import logging
import re

from .arch import Arch, xxx


l = logging.xxx
```

archinfo/arch_soot.py
```diff
@@ -40,13 +45,56 @@ def from_string(cls, tstr):
...
+        :return: True, if name of soot method matches the mangled native name.
+        """
+        
+        if "__" in native_method:
+            # if native methods are overloaded, two underscores are used
+            # TODO check function signature
+            raise NotImplementedError('Overloaded native methods are not supported.')
+
+        # demangle native name
+        native_method =  native_method.replace('_1', '_')
```
ltfish: There is an extra space after `=`.

archinfo/arch_soot.py
```diff
@@ -40,13 +45,56 @@ def from_string(cls, tstr):
...
+        # TODO unicode escaping
```
ltfish: Just curious, when do you see Unicode strings in method names?

archinfo/arch_soot.py
```diff
@@ -40,13 +45,56 @@ def from_string(cls, tstr):
...
+        return native_method == method_native_name
+
     @classmethod
-    def from_method(cls, method):
-        return cls(method.class_name, method.name, method.params)
+    def from_soot_method(cls, soot_method):
+        return cls(class_name=soot_method.class_name, 
+                   name=soot_method.name, 
+                   params=soot_method.params, 
+                   soot_method=soot_method)
```
ltfish: The right parenthesis should go to the next line.

```
    soot_method=soot_method,
    )
```

archinfo/arch_soot.py
```diff
@@ -1,12 +1,17 @@
 
 from .arch import Arch, register_arch, Endness
+import re
+
+import logging
+l = logging.getLogger('archinfo.arch_soot')
 
 
 class SootMethodDescriptor(object):
-    def __init__(self, class_name, name, params):
```
ltfish: A general thought: This is a good chance to add `__slots__` to all these descriptors so it's faster to create them in CPython (and saves some memory).

archinfo/arch_soot.py
```diff
@@ -123,25 +171,145 @@ def __repr__(self):
...
+    name = 'Soot'
 
     vex_arch = None  # No VEX support
     qemu_name = None  # No Qemu/Unicorn-engine support
     bits = 64
     address_types = (SootAddressDescriptor, )
     function_address_types = (SootMethodDescriptor, )
 
-    name = 'Soot'
+    # Size of native counterparts of primitive Java types
```
ltfish: Good job. Obviously I never bothered to define these...

archinfo/arch_soot.py
```diff
@@ -123,25 +171,145 @@ def __repr__(self):
...
+                'J' : 'long',
+                'F' : 'float',
+                'D' : 'double',
+                'V' : 'void'
+               } 
+
+    @staticmethod
+    def decode_type_signature(type_sig):
+
+        if type_sig == "":
```
ltfish: Use `if not type_sig:`.

archinfo/arch_soot.py
```diff
@@ -123,25 +171,145 @@ def __repr__(self):
...
+    @staticmethod
+    def decode_method_signature(method_sig):
+        match = re.search(r'\((.*)\)(.*)', method_sig)
+        
+        params = tuple(ArchSoot.decode_type_signature(param) 
+                       for param in re.findall(r'([\[]*[ZBCSIJFDV]|[\[]*L.+;)', match.group(1)))
+        ret = ArchSoot.decode_type_signature(match.group(2))
+
+        l.debug("Decoded method signature '%s' as params=%s and ret=%s"
+                "" % (method_sig, params, ret) )
```
ltfish: Do not format strings in `.debug()` calls. You should rewrite it as the following:

```
l.debug("Decoded method signature '%s' as params=%s and ret=%s",
        method_sig, params, ret
        )
```

archinfo/arch_soot.py
rhelmot: I... disagree with this

ltfish: PEP8? Or whatever coding style I've been following.

rhelmot: it's well known that following pep8 religiously is bad

like I see the benefit to doing it that way, you get to have the trailing comma like in dict/list literals, but it looks soooooooooo ugly to my eyes...

ltfish: _welp_

Fine. I don't really care. Nitpicks anyway.

archinfo/arch_soot.py
thrsten: Thats a good question, which I don't have an answer for. However, I didn't give it much thought and just added the comment to be complete (since there exist an escape sequence for unicode characters in the JNI specification)

Phat3:
The PR has been made already in the past, and it should be up to date now with the recent changes in the code.

PR #43, opened 2018-07-11
Fix minor things

tyb0807:
@rhelmot There's one more bug here. I didn't read the comment for PPC about the BP register...

The other commit is to auto generate register offsets from `registers_list`


archinfo/arch.py
```diff
@@ -199,37 +199,19 @@ def __init__(self, endness, instruction_endness=None):
...
             self.concretize_unique_registers = set(r.vex_offset for r in self.register_list if r.concretize_unique)
-#           print 'concretize_unique_registers'
-#           for i in self.concretize_unique_registers:
-#               print i
+
+            # Register offsets
+            self.ip_offset = self.registers['ip'][0]
+            self.sp_offset = self.registers['sp'][0]
+            self.bp_offset = self.registers['bp'][0]
+            self.lr_offset = self.registers.get('lr', (None, None))[0]
```
rhelmot: I would do this by adding attributes to the Register class - `instruction_pointer=True, stack_pointer=True, base_pointer=True, link_register=True`. Relying on these names seems sketchy to me.

rhelmot: However, there are a lot of other sketchy things going on in archinfo, and this seems like a pretty safe change, so I'll merge it for now.

tyb0807: What are other sketchy things? I can try to fix them! But I think since these names are used by other modules (`angr`, `cle`, `pyvex`, ...), they gotta be safe enough?

rhelmot: Yeah, I would prefer angr, cle etc didn't use `state.regs.sp` etc in order to get the stack pointer. Ideally `state.sp` would do `state.registers.load(state.arch.sp_offset, state.arch.bits)`. maybe archinfo should provide the size so we don't have to guess with state.arch.bits either?

All this stuff is sketchy - you run into it immediately when you try to implement some exotic architecture like AVR that has different sizes for different pointers, multiple memories, etc. I definitely can't just give you a list...

PR #44, opened 2018-08-31
Add s390x

mephi42:


mephi42:
Will fix:
```
Traceback (most recent call last):
  File "/home/travis/.virtualenvs/angr/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/angr-dev/angr-doc/tests/test_apidoc.py", line 29, in test_api_coverage
    raise Exception("The following modules are not represented in the api docs:\n" + '\n'.join(missing))
Exception: The following modules are not represented in the api docs:
archinfo.arch_s390x
```

mephi42:
Yes, I have a few patches for other parts of angr, but they are not fully functional yet.

mephi42:
Apparently there is a cyclic dependency between this and https://github.com/angr/angr-doc/pull/206, which I pushed to fix the documentation test. Any idea how to resolve this?

rhelmot:
We don’t have the CI set up in a way that’ll synchronize correctly across
repos when the branches are on someone else’s fork. This’ll be fixed at
some point...

I’ll look at this and merge it tomorrow.

subwire:
As rhelmot said, its a quirk of CI.  One of us will handle that for you.

Feel free to reach out to us on Slack if you need pointers on the other components, we're happy to see angr extended to cool new analysis targets

Particularly, if you have a toolchain set up, and wouldn't mind compiling some of our tests (in the `binaries`) repo, that would be awesome! That way, we can track it via CI. Some of our tests try all the arches, so you just need to add s390x to the lists and you're good to go there.

rhelmot:
Merged! Looks fantastic.

Some notes for myself/you going forward:

- the identifier in the list in register_arch is supposed to be a bunch of regexes that could match any string which refers to the arch, not just from the elf machine name enums
- I hadn't noticed that we still have the syscall and return value register offsets hardcoded. we're trying to move away from hardcoding those offsets because they can float around from compiler to compiler, so in the future we should add a flag indicating those semantics to the Register class
- we're very close to migrating to python 3 permanently, make sure all your bytestrings start with `b`` ;)
- I'm kind of horrified looking at the default behavior of the `Arch.__init__` base constructor which automatically tries reversing the endness on the hardcoded instructions - I guess we never had to deal with a big-endian architecture with variable width instructions. The end-reversing should probably be moved to the arch-specific initializers.

PR #45, opened 2018-09-05
arch_from_id: throw ArchNotFound on error

mephi42:
This change makes it possible for cle's check_compatibility() to
gracefully ignore unknown arches while still failing on real errors.


mephi42:
Fails in `ANGR_REPO=fidget` on all ARM tests. Not sure if this patch can really cause this? I see the same issue in an unrelated https://github.com/angr/cle/pull/147.

rhelmot:
Those tests broke at some point. I’ll have to bisect for what broke them
when I have time. Don’t worry about it for now.

PR #52, opened 2018-11-01
Add methods for address conversion between real/executable variants

danse-macabre:
See the discussion at https://github.com/angr/angr/pull/976


rhelmot:
I have some qualms about this in the long run but it's good for now. I suggest that we merge this and take the notion of "what is an address" into consideration when designing our new knowledge base.

PR #53, opened 2018-11-18
Allow translating sub-registers to base registers.

ltfish:


ltfish:
@rhelmot We mainly deal with (VEX) register offsets in angr (both in static analysis code and symbolic execution code). I wonder how else this can be designed without using VEX offsets and without incurring a lot of conversion from VEX offsets to register names and back to VEX offsets when using it.

rhelmot:
Right - but in the meantime, can we make it so that this doesn't cause crashes when pyvex isn't installed?

PR #55, opened 2019-01-07
Allow querying whether a register exists on a real CPU or not.

ltfish:


ltfish:
CI failed because of my changes in `ailment`. It will be resolved when all my angr and ailment PRs are merged in.

PR #59, opened 2019-02-18
ARM CFG party favor:  Cortex-M support

subwire:
This PR adds the following:

- Support for the ARMCortexM architecture.  This isn't your average ARM:
+ Supports only Thumb2 instructions!
+ New registers!
+ You really need those pesky Thumb prologues to build a CFG
+ The processor "state" (the CPU flags) are actually important, since we have no OS to do that stuff for us
+ New, magical FPU stuff

This adds a separate arch for CM, and the future PRs for other repos will enable its use.
All but the last two are now handled; QEMU doesn't even support that FPU nonsense yet, it's a real mess.
Prologues are now split into thumb and non-thumb ones, to help avoid false-positives from those short two-byte push xxx and pop xxx instructions.

### DIFF
--- a/archinfo/arch_aarch64.py
+++ b/archinfo/arch_aarch64.py
@@ -134,7 +134,7 @@ class ArchAArch64(Arch):
                  general_purpose=True),
         Register(name='x30', size=8, subregisters=[('w30', 0, 4)], alias_names=('r30', 'lr'),
                  general_purpose=True),
-        Register(name='xsp', size=8, subregisters=[('wsp', 0, 4)], alias_names=('sp',),
+        Register(name='xsp', size=8, subregisters=[('wsp', 0, 4)], alias_names=('sp','xsp   '),
                  general_purpose=True, default_value=(initial_sp, True, 'global')),
         Register(name='pc', size=8, alias_names=('ip',)),
         Register(name='cc_op', size=8, artificial=True),
--- a/archinfo/arch_aarch64.py
+++ b/archinfo/arch_aarch64.py
@@ -134,7 +134,7 @@ class ArchAArch64(Arch):
                  general_purpose=True),
         Register(name='x30', size=8, subregisters=[('w30', 0, 4)], alias_names=('r30', 'lr'),
                  general_purpose=True),
-        Register(name='xsp', size=8, subregisters=[('wsp', 0, 4)], alias_names=('sp','xsp   '),
+        Register(name='xsp', size=8, subregisters=[('wsp', 0, 4)], alias_names=('sp',),
                  general_purpose=True, default_value=(initial_sp, True, 'global')),
         Register(name='pc', size=8, alias_names=('ip',)),
         Register(name='cc_op', size=8, artificial=True),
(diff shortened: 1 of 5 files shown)


subwire:
@ltfish this is ready.  The remaining failures appear to be some temporal glitch in CI (the bugs are fixed on the angr side)
I think it's time we just do this already.

archinfo/arch_aarch64.py
```diff
@@ -134,7 +134,7 @@ def __init__(self, endness=Endness.LE):
                  general_purpose=True),
         Register(name='x30', size=8, subregisters=[('w30', 0, 4)], alias_names=('r30', 'lr'),
                  general_purpose=True),
-        Register(name='xsp', size=8, subregisters=[('wsp', 0, 4)], alias_names=('sp',),
+        Register(name='xsp', size=8, subregisters=[('wsp', 0, 4)], alias_names=('sp','xsp   '),
```
ltfish: Why the spaces in `'xsp    '`? Also, you missed a space in front of this string.

archinfo/arch_arm.py
```diff
@@ -37,7 +37,8 @@ def __init__(self, endness=Endness.LE):
         if endness == Endness.BE:
             self.function_prologs = {
                 br"\xe9\x2d[\x00-\xff][\x00-\xff]",          # stmfd sp!, {xxxxx}
-                br"\xe5\x2d\xe0\x04",                        # push {lr}
+                br"\xe5\x2d\xe0\x04",# push {lr}
```
ltfish: At least two spaces in front of an inline comment.

ltfish: Also I'm not sure why you are touching this line...

archinfo/arch_arm.py
```diff
@@ -274,7 +281,83 @@ class ArchARMEL(ArchARM):
...
+        # TODO: POP.W
+    }
+
+    def __init__(self, *args, **kwargs):
+        ArchARMEL.__init__(self, *args, **kwargs)
+
+        # See http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0553a/CHDBIBGJ.html
+        # for relevant pain-points.  Most of these only get updated on an exception/interrupt/syscall
+        # Except we leave T on because we have to.
+        self.register_list += [
```
rhelmot: Shouldn't this happen before the super-call to `__init__`? the base initializer does a lot of processing with the registers, and it should be where e.g. the register file allocation for non-vex registers happens (where are these assigned offsets, anyway..?)

archinfo/arch_arm.py
```diff
@@ -274,7 +281,83 @@ class ArchARMEL(ArchARM):
...
+        br"[\x00-\xff]\xb5",  # push {xxx,lr}
+        br"[\x00-\xff][\x00-\xff]\x2d\xe9", # push.w {xxx, lr}
+    }
+    function_epilogs = {
+        br"[\x00-\xff]\xbd" # pop {xxx, pc}
+        # TODO: POP.W
+    }
+
+    def __init__(self, *args, **kwargs):
+        ArchARMEL.__init__(self, *args, **kwargs)
```
rhelmot: This is old syntax, it should be `super(ArchARMEL, self).__init__` or `super().__init__` (the former is the full syntax, the latter is an abbreviated syntax introduced in some py3k minor version)

rhelmot: *`super(ArchARMCortexM, self).__init__`

subwire: Done

archinfo/arch_arm.py
subwire: Done

PR #63, opened 2019-04-02
Import without pyvex

### LINKED ISSUE #62: Confusing error message when pyvex fails to import
If `pyvex` fails to import, [archinfo/arch.py](https://github.com/angr/archinfo/blob/master/archinfo/arch.py) sets `_pyvex` to `None` (line 13). The rest of the code is very careful to check that `_pyvex` is not `None` before using it, except line 188, which references `_pyvex` without first checking that it is not `None`. As a result, a failed import of `pyvex` gives the error message `'AttributeError: 'NoneType' object has no attribute 'vex_ffi'` on line 188, which is a very confusing way to say "pyvex failed to import". I'm not sure whether the proper fix is to simply move the two lines 188-189 inside the if-block which begins at line 191? or if something more complicated is needed. If the former, I'm happy to put up a PR.

cdisselkoen:
Fixes #62.  Includes testcases for importing archinfo in various conditions.

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -184,11 +184,11 @@ class Arch:
             self.ret_instruction = reverse_ends(self.ret_instruction)
             self.nop_instruction = reverse_ends(self.nop_instruction)
 
-        if self.register_list:
+        if self.register_list and _pyvex is not None:
             (_, _), max_offset = max(_pyvex.vex_ffi.guest_offsets.items(), key=lambda x: x[1])
             max_offset += self.bits
             # Register collections
-            if type(self.vex_arch) is str and _pyvex is not None:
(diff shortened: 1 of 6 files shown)


ltfish:
Thanks! This PR looks good to me. Just one thing: All our test files are named like the following: `test_xxx.py`. We don't use camel case for Python files in this project. Can you please rename those test files accordingly?

cdisselkoen:
Done, no problem.

ltfish:
I'm not sure if those test scripts will be picked up if there are no `test_`-prefixed methods inside. I'm merging this PR in anyway since we don't have CI support for it at the moment. It will get fixed once we hook this up to our CI.

Thanks again!

PR #66, opened 2019-04-16
ARM Quality-of-life fixes and new register API

subwire:
This PR adds:

- Fix CMx *stone supoort
- Update ARM registers to reflect what's real and what's not.
This is not complete, as the issue of concrete vs synthetic status registers and components is a tricky one across many arches.
- Add a function get_register_by_name, a simple quickie for getting a Register based on its name, or any of its aliases.  This helps when you have environments (concrete vs symbolic) you want to synchronize, and their register files have slightly different naming.
We also therefore standardize that the registers named or aliased "ip", "sp", and "bp" have special meanings. We should use arch.get_register_by_name('sp'), etc instead of the nasty hacks we have around angr for this, particularly to remove dependence on VEX and its "offsets"

This is all designed to work with, in case that wasn't obvious, an overhaul of concrete target support to work with ARM and some particularly concrete targets, that is coming very soon

### DIFF
--- a/archinfo/arch_arm.py
+++ b/archinfo/arch_arm.py
@@ -229,15 +229,15 @@ class ArchARM(Arch):
         Register(name='cc_dep1', size=4, default_value=(0, False, None), artificial=True),
         Register(name='cc_dep2', size=4, default_value=(0, False, None), artificial=True),
         Register(name='cc_ndep', size=4, default_value=(0, False, None), artificial=True),
-        Register(name='qflag32', size=4, vector=True),
-        Register(name='geflag0', size=4, vector=True),
-        Register(name='geflag1', size=4, vector=True),
-        Register(name='geflag2', size=4, vector=True),
-        Register(name='geflag3', size=4, vector=True),
-        Register(name='emnote', size=4, vector=True),
-        Register(name='cmstart', size=4, artificial=True),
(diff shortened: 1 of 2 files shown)


archinfo/arch_arm.py
```diff
@@ -336,11 +336,32 @@ class ArchARMCortexM(ArchARMEL):
...
+    @property
+    def keystone_thumb(self):
+        return self.keystone
 
         # See http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0553a/CHDBIBGJ.html
         # for relevant pain-points.  Most of these only get updated on an exception/interrupt/syscall
         # Except we leave T on because we have to.
+
+    # EDG says: Note, due to the thing about class members being single objects, don't do this
+    # in the constructor, or you're gonna have a bad time
```
rhelmot: But you _are_ doing this in the constructor. Do you mean in the class body? You can just say `register_list = ArchARM.register_list + [...]`

subwire: @rhelmot but you cactually can't -- that's a class object, so each time you make an ArchARMCortexM() you'll add those to the list, and get lots of duplicates. 

I solved the problem by duplicating the list.  I also simplified it and made it more realistic while I'm at it.

PR #68, opened 2019-06-20
Add high quadword (xmm*hq) and low quadword (xmm*lq) subregisters of xmm*

saullocarvalho:
Fix issue angr/pyvex#194.


twizmwazin:
I'm assuming that formatting came out of an autoformatter? Is it possible to disable the line length restrictions for this section, the pre-patch version reads much cleaner.

saullocarvalho:
Should each register use only one line?
Should I use
```
Register(name='ymm0', size=32, subregisters=[('xmm0', 0, 16), ('xmm0lq', 0, 8), ('xmm0hq', 8, 8)], vector=True),
```
instead of
```
Register(name='ymm0', size=32, subregisters=[('xmm0', 0, 16),
                                             ('xmm0lq', 0, 8),
                                             ('xmm0hq', 8, 8)],
         vector=True),
```
?

twizmwazin:
That's what I would prefer, however I don't know others' preferences.

saullocarvalho:
I've used https://github.com/angr/archinfo/blob/75c62c8c40b6635e555c23e85777cb74ef880fd0/archinfo/arch_amd64.py#L148-L152 as a formatting reference.

twizmwazin:
Ah, I see. Consistency is more important then.

rhelmot:
There are a lot of conflicting opinions on how to do line continuation formatting. We've decided that it doesn't really matter and whoever's writing the code can make it pretty in whatever way they want it. My preference would have been to stick a newline after the `[` and have everything else indented one level past the start of the starting line, but I didn't write the code :)

PR #69, opened 2019-07-12
move bytes to getattr from dynamic property

zardus:
This speeds up `arch.bytes` access from being responsible for 5% of the runtime down to 0.

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -158,6 +158,8 @@ class Arch:
 
     def __init__(self, endness, instruction_endness=None):
 
+        self.bytes = self.bits // self.byte_width
+
         if endness not in (Endness.LE, Endness.BE, Endness.ME):
             raise ArchError('Must pass a valid endness: Endness.LE, Endness.BE, or Endness.ME')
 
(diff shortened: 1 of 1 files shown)


archinfo/arch.py
```diff
@@ -374,12 +374,14 @@ def _get_register_dict(self):
...
-        The standard word size in bytes, calculated from the ``bits`` field
-        """
-        return self.bits // self.byte_width
+    def __getattr__(self, a):
+        if a == 'bytes':
+            # The standard word size in bytes, calculated from the ``bits`` field
+            v = self.bits // self.byte_width
+        else:
+            raise AttributeError(a)
+        setattr(self, a, v)
```
Lukas-Dresel: Should we hook `__setattr__` of `bits/byte_width` so that they don't desync if any madman ever changes them after? If not why don't we just store the attribute directly? Then we don't even need the `__getattr__`

rhelmot:
You are now the third person to suggest this change. It’ll get moved into a
real attribute set in __init__

zardus:
@ConnorNelson , can you make a PR in angr-platforms to fix ArchBPF? We don't want to upset everyone by breaking that!

PR #70, opened 2019-07-23
Updated test_import.py

anupriya-gupta:


ltfish:
Why are you making this change?

anupriya-gupta:
Hello Sir, I am working as a GSA in your lab and I have been assigned the task of writing unit tests for archinfo by Prof Yan. So I had created this pull request.

ltfish:
> I am working as a GSA in your lab

That I'm aware ;)

> and I have been assigned the task of writing unit tests for archinfo by Prof Yan.

I'm also aware of that. He's better referred to as "Yan" or "Prof. Shoshitaishvili" though.

> So I had created this pull request.

Right, I'm just asking what the idea is behind this unit test. Do we expect those strings (`LE`, `BE`, and `ME`) to be unchanged (which I disagree with)? Why do we put `Register` tests in `test_import.py`, which (from the name) is clearly about testing if `archinfo` imports?

anupriya-gupta:
So I will remove this function test_endness() since there is no change on the coverage. I will also put the test_register() function in a separate file, so that test_import.py will only have tests related to importing archinfo.

tests/test_import.py
```diff
@@ -1 +1,4 @@
 import archinfo
+
+def test_import(): 
+    pass
```
twizmwazin: Would it make more sense to have the import here rather than module level?

tests/test_register.py
```diff
@@ -0,0 +1,26 @@
...
+                         vector = True, argument = True, persistent = True,
+                         default_value = (True, 'global'), linux_entry_value = 'argv',
+                         concretize_unique = True, concrete = True, artificial = True )
+    assert register.name == 'MIPS'
+    assert register.size == 8 
+    assert register.vex_offset == 10
+    assert register.vex_name == "xyz"
+    assert register.subregisters == ['a', 'b']
+    assert register.alias_names == ('r0')
+    assert register.general_purpose == True
```
twizmwazin: All these True asserts can be simplified to just test the value, this is why the CI is failing.

tests/test_import.py
twizmwazin: @anupriya-gupta Please follow up on this

PR #74, opened 2019-08-21
Tests for archsoot.py

anupriya-gupta:


tests/test_arch_soot.py
```diff
@@ -0,0 +1,430 @@
+from archinfo.arch import Endness
+from nose.tools import raises
+from archinfo.arch_soot import SootAddressDescriptor, SootMethodDescriptor, SootAddressTerminator, SootFieldDescriptor, SootClassDescriptor, SootNullConstant, SootArgument, ArchSoot
+import nose.tools
```
twizmwazin: These should probably be reordered. https://stackoverflow.com/questions/20762662/whats-the-correct-way-to-sort-python-import-x-and-from-x-import-y-statement might be helpful.

tests/test_import.py
```diff
@@ -1 +1,2 @@
-import archinfo
+def test_import():
```
twizmwazin: I think this is the wrong PR for this change

tests/test_register.py
```diff
@@ -0,0 +1,25 @@
+def test_register():
```
twizmwazin: I think this is the wrong PR for this change

tests/test_arch_soot.py
anupriya-gupta: Okay. I'll do that.

PR #75, opened 2019-10-15
Type annotations for archinfo

fmagin:
Discussion in angr/pyvex#198

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -5,7 +5,7 @@ import struct as _struct
 import platform as _platform
 import re
 from archinfo.archerror import ArchError
-from pyvex import RegisterOffset, RegisterName
+from archinfo import RegisterOffset, RegisterName
 
 import copy
 
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, List, Tuple, Any, Optional
+from typing import Dict, List, Tuple
 
 import struct as _struct
 import platform as _platform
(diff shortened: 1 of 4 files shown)


archinfo/arch.py
```diff
@@ -1,8 +1,12 @@
 import logging
+from typing import Dict, List, Tuple
+
 import struct as _struct
 import platform as _platform
 import re
 from archinfo.archerror import ArchError
+from pyvex import RegisterOffset, RegisterName
```
twizmwazin: We shouldn't be using pyvex in archinfo.

setup.py
```diff
@@ -15,4 +15,7 @@
     version='8.20.1.7',
     python_requires='>=3.5',
     packages=packages,
+    package_data={'archinfo': ['py.typed']},
+    test_suite='nose.collector',
+    tests_require=['nose'],
```
twizmwazin: I'm not sure that we want these two lines, at least as part of the PR.

fmagin: That was probably some rebase related messup and definitely not intentional, will fix.

archinfo/arch.py
fmagin: This raises the issue of where to put those type aliases. They need to be unique, so it is not possible for each module to define it themselves. The only other option that comes to mind right now would be a dedicated module for type aliases like this. Declaring that some integer is a RegisterOffset and not a TempVar is one of the most useful benefits of type annotations for this part of the codebase, so I definitely want this somewhere.
The meaning mapping registername <-> register offset comes from Vex it seems reasonable to me to get it from pyvex. What issues would be introduced by this?

rhelmot: archinfo just needs to still work when pyvex isn't installed, just without the pyvex features. you can put this in an optional idiom and make them None if pyvex throws an ImportError.

Alternately, the type aliases can just live in archinfo.

PR #77, opened 2019-11-15
Define how to run tests in `setup.py`

Pamplemousse:
`python setup.py test` is a conventional way to run tests, and now
works.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,4 +15,6 @@ setup(
     version='8.19.10.30',
     python_requires='>=3.5',
     packages=packages,
+    test_suite='nose.collector',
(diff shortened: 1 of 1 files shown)


rhelmot:
TIL. Thanks!

Pamplemousse:
Ah!
Used to be conventional, but got deprecated: pypa/setuptool#1878!

Sorry I did not realise.

Poke @twizmwazin .

rhelmot:
:(

I think `nosetests tests` can be an ok "default testing command". I have `alias nose='nosetests -v --nologcapture` so all I have to type is `nose tests`

PR #82, opened 2020-03-03
Use find(1) in a more portable way

0mp:
Not all find(1) implementation allow the user to skip
the directory on the command-line.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ try:
 except ImportError:
     from distutils.core import setup
     import os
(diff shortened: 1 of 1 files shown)


rhelmot:
Thanks. It is my hope that that line of code will never run and has never been run, since it's a fallback for if you don't have setuptools installed and if you don't have setuptools installed you're in for a seriously rough time, but it's nice to have things squared away.

0mp:
Haha, yeah, that piece of code looked very special. ;)

Cheers!

PR #84, opened 2020-04-27
Add RISC-V Architecture 

wwwzbwcom:
Hello, I add basic RISC-V information.

## Changes
- add information, register list about the RISC-V architecture (in `arch_riscv.py`, `ArchRISCV`)
- with `capstone >= 5.0.0` and `angr`, can disassemble RISC-V elf, binary etc.
- add simple import test for RISC-V (in `test_riscv.py`)
- add a new line in `.gitignore` to prevent adding build folder of `python setup.py`

## Know Issues
- `elf_tls` attribute is incorrect! (I know nothing about TLS)


rhelmot:
A risc-v vex lifter was contributed in https://github.com/angr/angr-platforms/pull/35. Could you redo this to be compatible with that PR? I genuinely don't know if the arch should live in archinfo or in angr-platforms, though i have a minor preference for angr-platforms because it's already there. I'm open to persuasion.

wwwzbwcom:
Thanks for your advice!  I redo that, now this is compatible with that PR.

I think the `ArchRISCV` should live in `archinfo`, according to this [official documentation](https://docs.angr.io/introductory-errata/helpwanted#additional-architectures). (In addition, most architecture class live here)

Also, if `ArchRISCV` should live in `angr-platform`, I can redo my work there, Thanks!

archinfo/arch_riscv.py
```diff
@@ -0,0 +1,164 @@
...
+    qemu_name = None  # No Unicorn-engine support
+    ida_processor = None
+    triplet = 'riscv32-linux-gnu'
+    max_inst_bytes = 4
+
+    ip_offset = None
+    sp_offset = None
+    bp_offset = None
+    lr_offset = None
+    ret_offset = 11  # a0(x10) For return value
```
rhelmot, line 34: the meaning of "offset" here is the vex offset into the register file. 11 is definitely not the offset of a register... same with `syscall_num_offset` below. Also at the very least you should definitely fill in `ip_offset`, if not the rest as well.

rhelmot:
I think I would like this to be a PR to angr-platforms, just to keep all the risc-v stuff in the same place.

wwwzbwcom:
ok, I will fix those issues and redo my work there. Thanks for your advice!

PR #86, opened 2020-05-06
Type aliasing

fmagin:
As discussed on Slack. Relates to

`NewType` [0] types for `RegisterOffset` and `TmpVar` should not be an issue because they should nearly never be used by just using an integer literal and instead should be the result of some function or from some datastructure anyway. If this function or datastructure is properly typed this will not cause type issues and this will typically only be required in very few places and then propagate.

RegisterName is a different issue because just writing the string `"rax"` or similar will still be fairly common  and having to write it as `RegisterName("rax")` to satisfy the type checker quickly becomes annoying and causes more issues than it solves. Instead Type Aliases[1] can be used, which still allow `RegisterName` to be used for annotating functions and datastructures, but `str` can be used interchangeably with `RegisterName` (because it is the same object anyway). If it is deemed preferable to use a NewType instead at any point in the future, it will be a simple one line change back to `RegisterName = NewType("Registername", str)`

[0] https://docs.python.org/3/library/typing.html#newtype
[1] https://docs.python.org/3/library/typing.html#type-aliases

### DIFF
--- a/archinfo/__init__.py
+++ b/archinfo/__init__.py
@@ -8,7 +8,13 @@ __version__ = (8, 20, 1, 7)
 if bytes is str:
     raise Exception("This module is designed for python 3 only. Please install an older version to use python 2.")
 
-# pylint:disable=wildcard-import
+# Type Aliases
+from typing import NewType
+RegisterOffset = NewType('RegisterOffset', int)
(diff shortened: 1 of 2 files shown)


ltfish:
Merged. Thank you!

PR #93, opened 2020-09-25
Mark ftop register as an artificial register on x86 and amd64

dnivra:
ftop register should be an artificial register on x86 and amd64: it's used to store index of the register that is currently the top of the FP stack.

### DIFF
--- a/archinfo/arch_amd64.py
+++ b/archinfo/arch_amd64.py
@@ -325,7 +325,7 @@ class ArchAMD64(Arch):
                                                       ('xmm15hq', 8, 8),
                                                       ('ymm15hx', 16, 16)],
                  vector=True),
-        Register(name='ftop', size=4, floating_point=True, default_value=(0, False, None)),
+        Register(name='ftop', size=4, floating_point=True, default_value=(0, False, None), artificial=True),
         Register(name='fpreg', size=64, subregisters=[('mm0', 0, 8),
                                                       ('mm1', 8, 8),
                                                       ('mm2', 16, 8),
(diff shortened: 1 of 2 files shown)


rhelmot:
Remind me what the artificial parameter affects? It's definitely not something which has no direct analogue in the x86 register file, there definitely has to be some sort of pointer which tells you with of the x87/mmx registers corresponds to the st0 name and etc...

dnivra:
I don't really know if anything else in the angr suite ecosystem uses `artificial` field: it is used in the WIP unicorn engine rewrite to decide if some state needs to be tracked for registers or not during concrete emulation.

From what I checked before sending in the PR, yes ftop doesn't have a corresponding register in x86. I dived in a bit more to understand FP stack usage better. The stack top in x86 FPU is stored as a field in the FP status register(FPSW; ref: https://xem.github.io/minix86/manual/intel-x86-and-64-manual-vol1/o_7281d5ea06a5b67a-194.html), which I noticed just now that it isn't declared as a register in archinfo. Most floating point instructions that modify the stack implicitly update the top field(https://wiki.osdev.org/FPU#Programming_the_FPU) and there are also explicit instructions that can modify the top value(eg: https://www.felixcloutier.com/x86/fdecstp). The top pointer of FP stack can also be modified by modifying the FPSW(not directly possible: copy to somewhere, modify and store back into FPSW).

I suppose VEX chose to define ftop as an additional register instead of having to refer back to FPSW each time and so it is an artificial register(given how registers are deemed artificial in archinfo: https://github.com/angr/archinfo/blob/master/archinfo/arch.py#L74). But, there is a pointer that tracks the index of the stack: it's not an explicit register but a field in the FPSW. So, it's value does exist in the FP CPU during execution. Now, I admit it is more confusing what is the right thing to do.

rhelmot:
let's call this good for now. the test failures look unrelated (famous last words...)

PR #98, opened 2021-09-18
x86/amd64: Fixed default value of FTOP

dnivra:
According to the Intel software developers manual, the floating point register stack grows towards offset 0. At start of execution, the stack would be empty and so default value of FTOP should be 7 instead of 0.


dnivra:
The failing test case seems to be unrelated to this fix.

ltfish:
Merged. Thanks!

PR #101, opened 2021-12-24
Moved x86/amd64 specific setup code to arch specific files

dnivra:
The register blacklist and CPU flags setup code is x86/amd64 specific and should not be in arch.py but in arch_x86.py/arch_amd64.py. I am not sure why CI runner 8 timed out: the last run test from angrop completed in 2 minutes locally.


ltfish:
I just restarted the CI.

ltfish:
Thanks!

PR #104, opened 2022-02-01
Change default values to make iteration play nice.

novafacing:
Fixes an issue when running Symbion on ARM architectures with Unicorn enabled:

```
File "/home/[...snip...]/angr/state_plugins/unicorn_engine.py", line 1259, in set_regs
    if r in self.state.arch.reg_blacklist:
TypeError: argument of type 'NoneType' is not iterable
```

[x86](https://github.com/angr/archinfo/blob/bd84450115db29bef89022e3c63c7ea84facd46e/archinfo/arch_x86.py#L57) and [x86_64](https://github.com/angr/archinfo/blob/bd84450115db29bef89022e3c63c7ea84facd46e/archinfo/arch_amd64.py#L68) already initialize these to empty list.

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -708,8 +708,8 @@ class Arch:
     artificial_registers_offsets = None
     artificial_registers = set()
     cpu_flag_register_offsets_and_bitmasks_map = None
-    reg_blacklist = None
-    reg_blacklist_offsets = None
+    reg_blacklist = []
+    reg_blacklist_offsets = []
     unicorn_flag_register = None
     vex_to_unicorn_map = None
     vex_cc_regs = None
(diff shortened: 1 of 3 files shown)


dnivra:
Seems okay to me and CI tests all pass. I think I set these to `None` simply following how the rest of unicorn related stuff are initialized. Along with this change, I think we can get rid of [lines 57 and 58 in archinfo/arch_x86.py](https://github.com/angr/archinfo/blob/bd84450115db29bef89022e3c63c7ea84facd46e/archinfo/arch_x86.py#L57) as well: no need to initialize them again there. What do you think @rhelmot?

rhelmot:
Seems okay to me as well. @novafacing, can you update your PR to remove the now-unnecessary initialization of these variables from the amd64/x86 arches?

novafacing:
Done!

rhelmot:
Thank!

PR #121, opened 2022-10-24
Fix import issues reported by pylint

twizmwazin:
This will probably actually make the lint decrease because of the de-duplicated code


twizmwazin:
I'm not sure I get the CI failures but I suspect it could be the result of import ordering shifting around.

ltfish:
It's likely that pyvex failed to import. Did you test it locally?

twizmwazin:
Testing locally pyvex does successfully import

PR #126, opened 2023-01-17
Add pre-commit

twizmwazin:
https://github.com/angr/pyvex/pull/292


mborgerson:
I've added the pre-commit script in #128 before I noticed this was open (oops). It's basically the same but doesn't include the cleanups here

twizmwazin:
Superseding with #129

PR #127, opened 2023-01-24
Add Arch generation via pypcode definitions

mborgerson:
This patch adds an ArchPcode architecture class for specific, minimal archinfo-compatible Archs generated from pypcode 'languages'. pypcode is a soft dependency.

Needs:
- [x] apidoc update
- [x] Fix unpickleable issue

Related:
* angr/cle#356
* angr/angr#3761
* angr/angr-management#879
* angr/angr-doc#450
* angr/binaries#102

### DIFF
--- a/tests/test_pcode.py
+++ b/tests/test_pcode.py
@@ -1,0 +1,31 @@
+# pylint:disable=missing-class-docstring,no-self-use
+import unittest
+import pickle
+
+try:
+    import pypcode
+except ImportError:
+    pypcode = None
+
+from archinfo.arch import Endness
+from archinfo import ArchPcode, ArchError
+
+
+@unittest.skipUnless(pypcode is not None, "pypcode not installed")
+class TestArchPcode(unittest.TestCase):
+    def test_arch_68000(self):
+        arch = ArchPcode("68000:BE:32:default")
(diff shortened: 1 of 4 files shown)


rhelmot:
To be clear, this is a strict dependency, right? pypcode has no knowledge that archinfo exists?

tests/test_pcode.py
```diff
@@ -0,0 +1,20 @@
+# pylint:disable=missing-class-docstring,no-self-use
+import unittest
+
+from archinfo.arch import Endness
+from archinfo import ArchPcode, ArchError
+
+
+class TestArchPcode(unittest.TestCase):
```
twizmwazin, line 8: Can we make sure this test is skipped if the soft dependency isn't installed?

mborgerson:
> To be clear, this is a strict dependency, right? pypcode has no knowledge that archinfo exists?

@rhelmot it's a soft dependency for archinfo, pypcode remains unaware of archinfo. If you don't have pypcode, archinfo will still work normally.

> Can we make sure this test is skipped if the soft dependency isn't installed?

@twizmwazin Good call, I'll skip the test if pypcode is not installed

PR #139, opened 2023-02-24
Fix duplicated arches in all_arches.

ltfish:


ltfish:
Not sure if you have this fixed in your plugins branch @rhelmot.

ltfish:
IMO we should also fix the equality check of `ArchARM` (then fix the boyscout test cases) so that it is not so bizarrely implemented. We should drop the current constraint `ArchARM will match with any ARM`. Please let me know what you think @rhelmot.

rhelmot:
I'm fine to drop that constraint

PR #162, opened 2023-06-28
Fix: Get Register object via subregister name

StackC00ki3:
The feature was mentioned in the comment, but never implemented.
https://github.com/angr/archinfo/blob/55ba856a57fc813913678a4920ada45d621595fa/archinfo/arch.py#L358


rhelmot:
Looks good. Thanks!

PR #177, opened 2023-08-28
Add enough types to pass mypy

twizmwazin:
This fixes up typing issues enough to pass mypy when run as follows:

`mypy --ignore-missing-import archinfo`

After this, we may want to consider implementing type stubs for capstone, keystone, and unicorn, as well as type more functions and/or use `--check-untyped-defs`. By default mypy does not check untyped functions. nor do they pass as of this PR.

Also todo: CI.

sync: https://github.com/angr/pyvex/pull/356


twizmwazin:
As far as I can tell, the largest flaw in how I've done this so far is that I tried to make Endness an actual type. I may revert that and just type it as `str` everywhere and make an issue about refactoring it into something real in the future.

rhelmot:
I don't think your current endness enum will actually work without 3.11's StrEnum - we have wayyyyyyyy too much legacy code that does direct comparisons with Iend_LE and friends

twizmwazin:
Yep. I saw StrEnum in the docs and figured that might be a good fit, but it's 3.11 and I don't see us bumping the min version for that. Doing it the way I did here would require a medium size refactor, mostly in angr proper (got cle and pyvex tests passing locally). We could consider backporting it on python<3.11.

rhelmot (approved):

PR #299, opened 2025-09-12
Disable unicorn from making a DeprecatedWarning print on the screen

mahaloz:
Unicorn will have the following Warning that prints on screen when importing:
```
UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
```

This disables that specific deprecation warning from printing.

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -3,8 +3,8 @@ import logging
 import platform as _platform
 import re
 import struct as _struct
-from typing import Any, Dict, List, Optional, Set, Tuple, Type, Union
 import warnings
+from typing import Any, Dict, List, Optional, Set, Tuple, Type, Union
 
 from archinfo.types import RegisterName, RegisterOffset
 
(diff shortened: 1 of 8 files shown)


twizmwazin:
By chance can those users just uninstall unicorn? We don't hard-depend on it anymore

mahaloz:
@twizmwazin is there a reason why this change is destructive? I'd rather do less relaying to people to change things and have it just update to be corrected. 

This stops a very niche type of signal from a package on its way out. I feel the benefits to the user outway the cons.

rhelmot:
this change is destructive because otherwise angr with unicorn is going to stop importing in december unless we do something about it

twizmwazin:
If there are other warnings you can point out, this is certainly an area we can improve on. Lots of warnings/info messages that could probably be debug messages. Ones warning of imminent compatibility errors, that would be difficult to debug if suppressed, doesn't seem like the right place to begin.

PR #349, opened 2026-06-01
AMD64: Set fp_ret_offset

mborgerson:
AMD64 SysV returns float/double in xmm0 (offset 224). Without this, the decompiler's AIL converter never created fp_ret_expr for call statements, so FP return values from calls were lost.

### DIFF
--- a/archinfo/arch_amd64.py
+++ b/archinfo/arch_amd64.py
@@ -152,6 +152,7 @@ class ArchAMD64(Arch):
     triplet = "x86_64-linux-gnu"
     max_inst_bytes = 15
     ret_offset = RegisterOffset(16)
+    fp_ret_offset = RegisterOffset(224)  # xmm0
     vex_conditional_helpers = True
     syscall_num_offset = 16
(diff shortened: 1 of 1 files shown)


rhelmot:
This is probably good to merge as-is, but the calling convention stuff in archinfo is a bit of a hackjob since it should really be handled on a per-calling-convention basis.

mborgerson:
> should really be handled on a per-calling-convention basis

Agree

ltfish:
> should really be handled on a per-calling-convention basis

It requires a much bigger refactor especially in the static analysis realm.

twizmwazin:
> It requires a much bigger refactor especially in the static analysis realm.

Can we replace the use of archinfo constants with the equivalent value from the default calling convention?

mborgerson:
I don't plan to address the architectural issues or refactoring in this PR, we'll handle it in a follow up

ltfish:
Can we just merge this PR? I think it's good enough as it is.

PR #353, opened 2026-06-23
Add capstone 6 compatibility for ARM64 and S390X architectures

kripticni:
Capstone 6 renamed CS_ARCH_ARM64 -> CS_ARCH_AARCH64 and CS_ARCH_SYSZ -> CS_ARCH_SYSTEMZ
in its Python bindings (old names removed entirely).

This adds try/except fallback so archinfo works with both capstone 5.x and 6.x.
The new constant name is tried first, with a fallback to the old name for older
capstone versions.

PR #354, opened 2026-06-25
Drop capstone <6.0 compat fallbacks and require capstone >=6.0

kripticni:
Drop capstone <6.0 compat fallbacks and require capstone >=6.0

Per maintainer @twizmwazin's instruction on PR #353:
"If bumping capstone to 6.0, please do remove the compat code for
capstone<6. It should remain an optional dependency."

Changes:
- arch_aarch64.py: remove hasattr() fallback, use CS_ARCH_AARCH64 directly
- arch_s390x.py: remove hasattr() fallback, use CS_ARCH_SYSTEMZ directly
- arch_riscv64.py: remove hasattr(CS_ARCH_RISCV) guard (exists in all >=6.0)
- pyproject.toml: add optional capstone>=6.0.0a1 dependency

Note: uses >=6.0.0a1 because capstone has not shipped a stable 6.0.0
on PyPI yet (latest is 6.0.0a9). PEP 440 correctly orders pre-releases
before the release, so >=6.0.0 would exclude all published versions.

### DIFF
--- a/archinfo/arch_aarch64.py
+++ b/archinfo/arch_aarch64.py
@@ -48,10 +48,7 @@ class ArchAArch64(Arch):
     instruction_endness = Endness.LE
     sizeof = {"short": 16, "int": 32, "long": 64, "long long": 64}
     if _capstone:
-        if hasattr(_capstone, "CS_ARCH_AARCH64"):
-            cs_arch = _capstone.CS_ARCH_AARCH64
-        else:
-            cs_arch = _capstone.CS_ARCH_ARM64
+        cs_arch = _capstone.CS_ARCH_AARCH64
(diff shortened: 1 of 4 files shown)


kripticni:
The ci job failures are expected. The angr/ci:3 container has capstone 5.x and the build installs angr@master which pins capstone==5.0.6. This PR needs the corresponding angr capstone bump at the same time.

The test job (archinfo pytest) passes fine as-is.

twizmwazin:
Oh, I did not realize 6.0 was not stable yet! Let's keep the compat code around so that the latest stable version is still supported. But don't close this PR, we can merge it once 6.0 is released as stable.

PR #363, opened 2026-08-09
Support narrow p-code stack pointers

zardus:

`ArchPcode` derived the initial stack pointer with an expression that assumed a stack pointer at least 16 bits wide, so constructing an architecture with a narrower one — the 8051 and the 8048 have a one-byte stack pointer on a 16-bit address space — raised `ValueError: negative shift count`.

Deriving the value from the stack pointer's own width instead fixes it where the register sizes are read, and leaves every architecture with a wider stack pointer exactly where it was.

The regression constructs `ArchPcode("8051:BE:16:default")` and pins the unchanged value for z80, 68000 and x86-64.

Validation: https://github.com/angr/archinfo/pull/363#issuecomment-5232368960. Related: angr/cle#717, angr/angr#6793.

session: mega-corpus

### DIFF
--- a/archinfo/arch_pcode.py
+++ b/archinfo/arch_pcode.py
@@ -122,7 +122,9 @@ class ArchPcode(Arch):
         sp_bits = self.bits
         if "sp" in archinfo_regs:
             sp_bits = archinfo_regs["sp"].size * 8
-        self.initial_sp = (0x8000 << (sp_bits - 16)) - 1
+        # start the stack at the top of the lower half of the stack pointer's range. The stack pointer can be
+        # narrower than the address space; the 8051, for instance, has a one-byte SP on a 16-bit address space.
+        self.initial_sp = (1 << (sp_bits - 1)) - 1
         self.linux_name = ""  # FIXME
         self.triplet = ""  # FIXME
 
--- a/tests/test_pcode.py
+++ b/tests/test_pcode.py
@@ -32,6 +32,17 @@ class TestArchPcode(unittest.TestCase):
         assert arch.registers["pc"] == arch.registers["rip"]
         assert arch.ip_offset == arch.registers["rip"][0]
 
+    def test_arch_with_narrow_stack_pointer(self):
+        """The 8051's stack pointer is one byte wide, narrower than its 16-bit address space."""
+        arch = ArchPcode("8051:BE:16:default")
+        assert arch.registers["sp"][1] == 1
+        assert arch.initial_sp == 0x7F
+
+        # architectures with a stack pointer at least 16 bits wide keep the initial stack pointer they always had
+        assert ArchPcode("z80:LE:16:default").initial_sp == 0x7FFF
+        assert ArchPcode("68000:BE:32:default").initial_sp == 0x7FFFFFFF
+        assert ArchPcode("x86:LE:64:default").initial_sp == 0x7FFFFFFFFFFFFFFF
+
     def test_arch_bad_langid(self):
         with self.assertRaises(ArchError):
             ArchPcode("invalid")


zardus:

Validation record for head `bfb1ced01e4b3cf760e37aa13232b9067fc653aa` against baseline `d17a90aecbf8450681086d18585e30a88d1cc7ad`. Rebased onto current master, which had gained two tests at the same point in `TestArchPcode` as the one added here; both sides are kept. `git range-diff` against the pre-rebase range shows the production change is byte-identical and only the test's insertion point moved, so the per-file lint and type scores below are carried forward from that head rather than re-measured.

- Reproducer: `python -c "import archinfo; archinfo.ArchPcode('8051:BE:16:default')"` — `ValueError: negative shift count` on the baseline
- Regression: `python -m pytest tests/test_pcode.py::TestArchPcode::test_arch_with_narrow_stack_pointer` — fails on the baseline with that error, passes on the head
- Full suite: `python -m pytest tests` — 20 passed on the rebased head
- Corpus: a sweep of this corpus recorded 1,330 objects failing with exactly this `ValueError`, across the 8051, 80390, 8048 and m8c families; all are p-code units, and none of them reaches any analysis, since the failure is in the architecture constructor
- Lint/type: pylint 4.0.6 and pyright 1.1.411, scored per changed file against the merge base the way the hosted Lint and Typecheck jobs score them — `archinfo/arch_pcode.py` 10.00 -> 10.00 and badness 0.238 -> 0.235, `tests/test_pcode.py` 10.00 -> 10.00 and badness 0.0 -> 0.0
- Hooks: `pre-commit run --all-files` — every hook passes and leaves the tree unchanged
- Coverage: on the head, constructing `ArchPcode` for every language pypcode 4.0.1 exposes succeeds, with no exception anywhere in the list
- Environment: Python 3.12.13, pytest 9.1.1, pypcode 4.0.1

Caveats: `initial_sp` is still taken from the address space width when the stack pointer register is not literally named `sp` (PIC-16 calls it `stkptr`), and `AARCH64:BE:32:ilp32` still gets a value from its 64-bit `sp` that sits outside the 32-bit address space; both predate this change and are untouched by it.

---

## Corpus measurement of the open queue, 2026-08-15 — this change only pays off with three others

The open pull-request queue was scored against 733 objects drawn from a sweep's own failing units (35 error classes, 49 architectures, 16 containers), with each repository's **current master** as the baseline rather than the revisions the sweep pinned. Each object is loaded with `auto_load_libs=False, use_sim_procedures=False` and then run through `CFGFast(normalize=True, data_references=False, resolve_indirect_jumps=True, force_complete_scan=False)` with a 120-second timeout.

Four of the narrow-p-code-architecture changes form a chain. Applied alone, each one clears part of its own class and leaves the rest of it standing on the next change's defect:

| Applied alone | Its own failure class | Objects that complete `CFGFast` | Where the remainder lands |
| --- | --- | ---: | --- |
| archinfo#363 | `ValueError: negative shift count` constructing `ArchPcode` — 1,673 units | 18 of 30 | `RecursionError` from state creation — angr#6807's defect |
| angr#6807 | `RecursionError` from state creation — 1,125 units | 20 of 30 | `KeyError: 24` at `lift` — angr#6793's defect |
| angr#6793 | `KeyError: 24` at `lift` — 1,423 units | 20 of 30 | `struct.error` in `pack_word` — cle#721's defect |
| cle#721 | `struct.error` in `pack_word` — 772 units | 1 of 24 | `KeyError: 24` at `lift` — back to angr#6793 |

Applied together — angr#6793 + cle#721 + cle#717 + archinfo#363 — the `KeyError: 24` class goes from 20 of 30 to **30 of 30**. Under the whole open queue all four classes are complete: 30 of 30, 30 of 30, 30 of 30 and 24 of 24, with no object lost in any class. Merging any one of them on its own therefore moves a fraction of the corpus units its own validation record describes, and the remainder only looks like a new failure.

One qualification about cle#717, which is included in the combination above: since its rebase onto cle#735 it carries no production change, only regression coverage, so nothing in the movement here is attributable to it. It guards the placement rule the other three depend on rather than supplying it.

For this PR specifically: the 30 measured objects are ones master cannot even build an architecture for, so they fail before any analysis starts. 18 of them complete `CFGFast` with this change alone. The other 12 reach state creation and die there, which is angr#6807. None of this changes the no-harm evidence above — every architecture with a wider stack pointer keeps its old `initial_sp`.




**Rebased 2026-08-27.** This record was measured at `bfb1ced01e4b3cf760e37aa13232b9067fc653aa`. The branch has since been rebased onto master at `72fcaa0b5750eebe6e88d219c66e5dceeda4d176`; `git range-diff` reports the commit unchanged, so only the base moved and every figure above still describes this head.

**Head currency, checked 2026-08-27.** This record was measured at `bfb1ced01e4b3cf760e37aa13232b9067fc653aa`. The branch has since been rebased onto master at `72fcaa0b5750eebe6e88d219c66e5dceeda4d176`; `git range-diff` over the two ranges reports every commit unchanged, so only the base moved and the figures above still describe this head.

PR #364, opened 2026-08-09
Reject word sizes struct cannot express

zardus:
THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

### Problem

`Arch.struct_fmt()` returns the format character `Z` for a three-byte word, and `struct` has no such character, so every caller that uses the string it hands back raises. It fires on the 14 of pypcode's 187 sleigh languages whose word is 24 bits — the PIC, dsPIC, HCS-12 and extended AVR families — and on any explicit `struct_fmt(size=3)`:

```
ArchPcode('HCS-12:BE:24:default')  bits=24 bytes=3
    .struct_fmt()                          -> '>Z'
    struct.calcsize('>Z')                  -> error: bad char in struct format
    struct.unpack('>Z', b'\x01\x02\x03')   -> error: bad char in struct format
```

The exception surfaces at the caller's `struct` call, naming a format string the caller never wrote, rather than at the architecture that produced it.

### Root cause

`archinfo/arch.py` maps size 3 to a character `struct` does not define:

```python
elif size == 3:
    fmt_size = "Z"  # special case for 24-bit architectures like AVR8
```

`struct` has integer format characters for 1, 2, 4 and 8 bytes and nothing for 3, so no character can be correct here. The `else` branch of the same chain already raises for 5, 6, 7 and 16; three bytes was the one invalid size that returned a value instead of raising.

### Fix

Delete the `size == 3` branch so a three-byte word falls into the existing `else`:

```
ArchPcode('HCS-12:BE:24:default')  bits=24 bytes=3
    .struct_fmt()                          -> ValueError: Invalid size: struct has no format character for a 3-byte integer
```

The message names the size rather than claiming the argument must be a power of two, which the old text asserted and which this function never required. A caller that has to read a 24-bit word assembles it from bytes; `struct_fmt` cannot do that for them, and the call is where saying so belongs. The docstring records the constraint so the next reader does not put a character back.

### Testing

`tests/test_struct_fmt.py` asserts that `struct.calcsize(arch.struct_fmt(size=n))` equals `n` for every size `struct` can express, signed and unsigned, and that 0, 3, 5, 6, 7 and 16 raise. `tests/test_pcode.py::test_struct_fmt_24bit` pins the p-code case. Both fail on master, where the 24-bit call returns `'>Z'` and `calcsize` rejects it.

angr/cle#721 is the other half of this. Validation: https://github.com/angr/archinfo/pull/364#issuecomment-5233538253

sync: angr/cle#721

session: sharpen

PR #365, opened 2026-08-09
Resolve sleigh language ids from arch_from_id

zardus:

### Problem

`arch_from_id` cannot name any of the 187 sleigh languages pypcode ships, so nothing that stores an architecture by name can read a p-code one back. angr's database serializer writes `arch.name` and reloads it through `arch_from_id`, as does `arch=` on `Project`.

```
arch_from_id('pa-risc:BE:32:default')  -> ArchNotFound: Can't find architecture info for architecture pa-risc:be:32:default with 32 bits and unsure endness
arch_from_id('ARM:LE:32:v7')  -> ArchARMEL(name='ARMEL', bits=32, memory_endness=<Endness.LE: 'Iend_LE'>)
```

Over all 187 languages, none round-trips: 117 raise, and the other 70 come back as the VEX architecture of the same processor. Those 70 are the worse ones — `ARM:LE:32:v7` reloads as `ArchARMEL`, a different lifter and a different register file, with no error anywhere.

### Root cause

`arch_from_id` resolves only what `register_arch` put in `arch_id_map`, and `ArchPcode` cannot be in it: `register_arch` binds a regex to a class constructed from an endness, while a p-code architecture is constructed from a sleigh language. A language id therefore reaches the regex table, where 78 of the 187 ids match a registered architecture's pattern for the same processor and the other 109 match nothing.

### Fix

`arch_from_id` moves out of `archinfo/arch.py` and into `archinfo/__init__.py`, whole, above the `__all__` list. An identifier containing `:` is tried as a language id first; everything else takes the path it always did, byte for byte. An id names exactly one language, so answering before `arch_id_map` is what stops `ARM:LE:32:v7` resolving to `ArchARMEL`; the `endness` and `bits` hints do not apply, because the id carries both.

```
arch_from_id('ARM:LE:32:v7')  -> ArchPcode(name='ARM:LE:32:v7', bits=32, memory_endness=<Endness.LE: 'Iend_LE'>)
```

`__init__.py` is the layer that can hold this: it already imports `.arch` and then `.arch_pcode` at module level, while `archinfo/arch.py` cannot import `ArchPcode` at the top of the file at all, because `arch_pcode` imports `arch`. So nothing this change adds needs a deferred import. `get_host_arch` moves with it — it is one call to `arch_from_id`, and leaving it in `arch.py` would have needed exactly the deferred import this removes.

`ArchError` from a non-matching id is suppressed, so a language-shaped identifier that names no language still falls through to the regex table. That covers a missing dependency too, since `ArchPcode.__init__` raises `ArchError("pypcode not installed")`, so `archinfo` without the `pcode` extra answers exactly as it did before.

One repair comes with the move. `register_arch`'s type check read `re._pattern_type`, which Python removed in 3.7. A list of plain strings short-circuits before reaching it, which is why archinfo's own 21 registrations work; any list holding something else raised `AttributeError`, both the compiled patterns the docstring documents and the bad types the check exists to reject. It is `re.Pattern`, and the annotation and the docstring now say what the check does. It is in this branch because `arch.py`'s diff is otherwise pure deletion, and pylint scores a file by messages per statement: without a compensating repair `ci / Lint` fails the file at 9.70 -> 9.66 with nothing added to it. With it, 9.70 -> 9.77.

`arch_from_id` and `get_host_arch` are no longer attributes of the `archinfo.arch` submodule. Nothing in angr, cle, pyvex, claripy, pypcode or angr-management imports either from there; every caller uses `archinfo.arch_from_id`.

### Testing

`tests/test_pcode.py` gains six cases: a language id no registered architecture can name resolves, and its own endness and width beat contradicting hints; four ids `arch_id_map` also claims resolve to `ArchPcode`; all 187 languages round-trip through `arch_from_id(arch.name, arch.memory_endness, arch.bits)`; ten registered names still resolve to their own classes; a language-shaped id naming nothing raises `ArchNotFound`; and, in a class that is not itself gated on pypcode being installed, `arch_from_id` answers exactly as before with `arch_pcode._has_pypcode` forced false, standing in for archinfo installed without the `pcode` extra. Three of the six fail on master, the round-trip one at the first language it enumerates.

`tests/test_register_arch.py` is new and covers the `re.Pattern` repair: a compiled regex registers and then resolves, and a list of the wrong type still raises `TypeError`. Both fail on master with the `AttributeError`.

Validation: https://github.com/angr/archinfo/pull/365#issuecomment-5233536803

session: sharpen

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -12,6 +12,7 @@ from .tls import TLSArchInfo
 from .types import Endness
 
 if TYPE_CHECKING:
+    # Typing only. arch_pcode imports this module, so its runtime import is delayed to each use site.
     from .arch_pcode import ArchPcode
 
 
@@ -881,7 +882,7 @@ def arch_from_id(ident: str, endness: str = Endness.ANY, bits: str | int = "") -
     """
     if _pypcode is not None and ":" in ident:
         # A language id names one language, so it answers before arch_id_map, whose regexes would otherwise
-        # claim ARM:LE:32:v7 for ArchARMEL. arch_pcode imports this module, so it cannot be imported at the top.
+        # claim ARM:LE:32:v7 for ArchARMEL. Delayed import to avoid circular dependency, as in Arch.pcode_arch.
         from .arch_pcode import ArchPcode  # pylint: disable=import-outside-toplevel
 
         with contextlib.suppress(ArchError):
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -1,3 +1,4 @@
+import contextlib
 import copy
 import logging
 import platform as _platform
@@ -863,13 +864,24 @@ class ArchNotFound(Exception):
     pass
 
 
-def arch_from_id(ident: str, endness=Endness.ANY, bits: str | int = "") -> Arch:
+def arch_from_id(ident: str, endness: str = Endness.ANY, bits: str | int = "") -> Arch:
     """
     Take our best guess at the arch referred to by the given identifier, and return an instance of its class.
 
     You may optionally provide the ``endness`` and ``bits`` parameters to help this function out. ``bits`` is
     either a number of bits or a string containing one, which is what an ELF class is.
+
+    A full sleigh language id, such as ``pa-risc:BE:32:default``, returns the ArchPcode for that language. It
+    carries its own endness and width, so the ``endness`` and ``bits`` hints do not apply to it.
     """
+    if ":" in ident:
+        # A language id names one language, so it answers before arch_id_map, whose regexes would otherwise
+        # claim ARM:LE:32:v7 for ArchARMEL. The import is delayed because arch_pcode imports this module.
+        from .arch_pcode import ArchPcode  # pylint: disable=import-outside-toplevel
+
+        with contextlib.suppress(ArchError):
+            return ArchPcode(ident)
+
     if bits == 64 or (isinstance(bits, str) and "64" in bits):
         bits = 64
     elif isinstance(bits, str) and "32" in bits:
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -1,4 +1,3 @@
-import contextlib
 import copy
 import logging
 import platform as _platform
@@ -12,7 +11,6 @@ from .tls import TLSArchInfo
 from .types import Endness
 
 if TYPE_CHECKING:
-    # Typing only. arch_pcode imports this module, so its runtime import is delayed to each use site.
     from .arch_pcode import ArchPcode
 
 
@@ -38,11 +36,6 @@ try:
     import keystone as _keystone
 except ImportError:
     _keystone = None
-
-try:
-    import pypcode as _pypcode
-except ImportError:
-    _pypcode = None
 
 
 class Register:
@@ -876,18 +869,7 @@ def arch_from_id(ident: str, endness: str = Endness.ANY, bits: str | int = "") -
 
     You may optionally provide the ``endness`` and ``bits`` parameters to help this function out. ``bits`` is
     either a number of bits or a string containing one, which is what an ELF class is.
-
-    A full sleigh language id, such as ``pa-risc:BE:32:default``, returns the ArchPcode for that language. It
-    carries its own endness and width, so the ``endness`` and ``bits`` hints do not apply to it.
     """
-    if _pypcode is not None and ":" in ident:
-        # A language id names one language, so it answers before arch_id_map, whose regexes would otherwise
-        # claim ARM:LE:32:v7 for ArchARMEL. Delayed import to avoid circular dependency, as in Arch.pcode_arch.
-        from .arch_pcode import ArchPcode  # pylint: disable=import-outside-toplevel
-
-        with contextlib.suppress(ArchError):
-            return ArchPcode(ident)
-
     if bits == 64 or (isinstance(bits, str) and "64" in bits):
         bits = 64
     elif isinstance(bits, str) and "32" in bits:
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -38,6 +38,11 @@ try:
 except ImportError:
     _keystone = None
 
+try:
+    import pypcode as _pypcode
+except ImportError:
+    _pypcode = None
+
 
 class Register:
     """
@@ -874,9 +879,9 @@ def arch_from_id(ident: str, endness: str = Endness.ANY, bits: str | int = "") -
     A full sleigh language id, such as ``pa-risc:BE:32:default``, returns the ArchPcode for that language. It
     carries its own endness and width, so the ``endness`` and ``bits`` hints do not apply to it.
     """
-    if ":" in ident:
+    if _pypcode is not None and ":" in ident:
         # A language id names one language, so it answers before arch_id_map, whose regexes would otherwise
-        # claim ARM:LE:32:v7 for ArchARMEL. The import is delayed because arch_pcode imports this module.
+        # claim ARM:LE:32:v7 for ArchARMEL. arch_pcode imports this module, so it cannot be imported at the top.
         from .arch_pcode import ArchPcode  # pylint: disable=import-outside-toplevel
 
         with contextlib.suppress(ArchError):
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -817,14 +817,14 @@ def _append_arch_unique(my_arch: Arch) -> bool:
     return True
 
 
-def register_arch(regexes: List[str], bits: int, endness: Endness, my_arch: Type[Arch]):
+def register_arch(regexes: List[Union[str, re.Pattern]], bits: int, endness: Endness, my_arch: Type[Arch]):
     """
     Register a new architecture.
     Architectures are loaded by their string name using ``arch_from_id()``, and
     this defines the mapping it uses to figure it out.
     Takes a list of regular expressions, and an Arch class as input.
 
-    :param regexes: List of regular expressions (str or SRE_Pattern)
+    :param regexes: List of regular expressions (str or re.Pattern)
     :type regexes: list
     :param bits: The canonical "bits" of this architecture, ex. 32 or 64
     :type bits: int
@@ -837,7 +837,7 @@ def register_arch(regexes: List[str], bits: int, endness: Endness, my_arch: Type
     if not isinstance(regexes, list):
         raise TypeError("regexes must be a list")
     for rx in regexes:
-        if not isinstance(rx, str) and not isinstance(rx, re._pattern_type):
+        if not isinstance(rx, str) and not isinstance(rx, re.Pattern):
             raise TypeError("Each regex must be a string or compiled regular expression")
         try:
             re.compile(rx)
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -1,6 +1,5 @@
 import copy
 import logging
-import platform as _platform
 import re
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple, Type, Union
 
@@ -863,73 +862,6 @@ class ArchNotFound(Exception):
     pass
 
 
-def arch_from_id(ident: str, endness: str = Endness.ANY, bits: str | int = "") -> Arch:
-    """
-    Take our best guess at the arch referred to by the given identifier, and return an instance of its class.
-
-    You may optionally provide the ``endness`` and ``bits`` parameters to help this function out. ``bits`` is
-    either a number of bits or a string containing one, which is what an ELF class is.
-    """
-    if bits == 64 or (isinstance(bits, str) and "64" in bits):
-        bits = 64
-    elif isinstance(bits, str) and "32" in bits:
-        bits = 32
-    elif not bits and "64" in ident:
-        bits = 64
-    elif not bits and "32" in ident:
-        bits = 32
-
-    endness = endness.lower()
-    if "lit" in endness:
-        endness = Endness.LE
-    elif "big" in endness:
-        endness = Endness.BE
-    elif "lsb" in endness:
-        endness = Endness.LE
-    elif "msb" in endness:
-        endness = Endness.BE
-    elif "le" in endness:
-        endness = Endness.LE
-    elif "be" in endness:
-        endness = Endness.BE
-    elif "l" in endness:
-        endness = Endness.UNSURE
-    elif "b" in endness:
-        endness = Endness.UNSURE
-    else:
-        endness = Endness.UNSURE
-    ident = ident.lower()
-    cls = None
-    aendness = None
-    for arxs, abits, aendness, acls in arch_id_map:
-        found_it = False
-        for rx in arxs:
-            if re.search(rx, ident):
-                found_it = True
-                break
-        if not found_it:
-            continue
-        if bits and bits != abits:
-            continue
-        if aendness == Endness.ANY or endness == aendness or endness == Endness.UNSURE:
-            cls = acls
-            break
-    if not cls:
-        raise ArchNotFound(
-            f"Can't find architecture info for architecture {ident} with {repr(bits)} bits and {endness} endness"
-        )
-    if endness == Endness.UNSURE:
-        if aendness == Endness.ANY:
-            # We really don't care, use default
-            return cls(cls.default_endness)
-        else:
-            # We're expecting the ident to pick the endness.
-            # ex. 'armeb' means obviously this is Iend_BE
-            return cls(aendness)
-    else:
-        return cls(endness)
-
-
 def reverse_ends(string: bytes) -> bytes:
     """
     Swap the endness of every four-byte word in ``string``.
@@ -939,10 +871,3 @@ def reverse_ends(string: bytes) -> bytes:
     """
 
     return b"".join(string[offset : offset + 4][::-1] for offset in range(0, len(string), 4))
-
-
-def get_host_arch():
-    """
-    Return the arch of the machine we are currently running on.
-    """
-    return arch_from_id(_platform.machine())
--- a/archinfo/__init__.py
+++ b/archinfo/__init__.py
@@ -7,16 +7,18 @@ It is useful for cross-architecture tools (such as pyvex).
 __version__ = "9.3.5.dev0"
 
 
+import contextlib
+
 from .arch import (
     Arch,
     ArchNotFound,
     Register,
     all_arches,
-    arch_from_id,
     get_host_arch,
     register_arch,
     reverse_ends,
 )
+from .arch import arch_from_id as _arch_from_id
 from .arch_aarch64 import ArchAArch64
 from .arch_amd64 import ArchAMD64
 from .arch_arm import ArchARM, ArchARMCortexM, ArchARMEL, ArchARMHF
@@ -63,3 +65,22 @@ __all__ = [
     "register_arch",
     "reverse_ends",
 ]
+
+
+def arch_from_id(ident: str, endness: str = Endness.ANY, bits: str | int = "") -> Arch:
+    """
+    Take our best guess at the arch referred to by the given identifier, and return an instance of its class.
+
+    You may optionally provide the ``endness`` and ``bits`` parameters to help this function out. ``bits`` is
+    either a number of bits or a string containing one, which is what an ELF class is.
+
+    A full sleigh language id, such as ``pa-risc:BE:32:default``, returns the ArchPcode for that language. It
+    carries its own endness and width, so the ``endness`` and ``bits`` hints do not apply to it.
+    """
+    # A language id names one language, so it answers before the registered architectures, whose regexes
+    # would otherwise claim ARM:LE:32:v7 for ArchARMEL.
+    if ":" in ident:
+        with contextlib.suppress(ArchError):
+            return ArchPcode(ident)
+
+    return _arch_from_id(ident, endness, bits)
--- a/archinfo/__init__.py
+++ b/archinfo/__init__.py
@@ -8,17 +8,18 @@ __version__ = "9.3.5.dev0"
 
 
 import contextlib
+import platform as _platform
+import re
 
 from .arch import (
     Arch,
     ArchNotFound,
     Register,
     all_arches,
-    get_host_arch,
+    arch_id_map,
     register_arch,
     reverse_ends,
 )
-from .arch import arch_from_id as _arch_from_id
 from .arch_aarch64 import ArchAArch64
 from .arch_amd64 import ArchAMD64
 from .arch_arm import ArchARM, ArchARMCortexM, ArchARMEL, ArchARMHF
@@ -33,6 +34,90 @@ from .arch_soot import ArchSoot
 from .arch_x86 import ArchX86
 from .archerror import ArchError
 from .types import Endness, RegisterName, RegisterOffset, TmpVar
+
+
+def arch_from_id(ident: str, endness: str = Endness.ANY, bits: str | int = "") -> Arch:
+    """
+    Take our best guess at the arch referred to by the given identifier, and return an instance of its class.
+
+    You may optionally provide the ``endness`` and ``bits`` parameters to help this function out. ``bits`` is
+    either a number of bits or a string containing one, which is what an ELF class is.
+
+    A full sleigh language id, such as ``pa-risc:BE:32:default``, returns the ArchPcode for that language. It
+    carries its own endness and width, so the ``endness`` and ``bits`` hints do not apply to it.
+    """
+    # A language id names one language, so it answers before the registered architectures, whose regexes
+    # would otherwise claim ARM:LE:32:v7 for ArchARMEL.
+    if ":" in ident:
+        with contextlib.suppress(ArchError):
+            return ArchPcode(ident)
+
+    if bits == 64 or (isinstance(bits, str) and "64" in bits):
+        bits = 64
+    elif isinstance(bits, str) and "32" in bits:
+        bits = 32
+    elif not bits and "64" in ident:
+        bits = 64
+    elif not bits and "32" in ident:
+        bits = 32
+
+    endness = endness.lower()
+    if "lit" in endness:
+        endness = Endness.LE
+    elif "big" in endness:
+        endness = Endness.BE
+    elif "lsb" in endness:
+        endness = Endness.LE
+    elif "msb" in endness:
+        endness = Endness.BE
+    elif "le" in endness:
+        endness = Endness.LE
+    elif "be" in endness:
+        endness = Endness.BE
+    elif "l" in endness:
+        endness = Endness.UNSURE
+    elif "b" in endness:
+        endness = Endness.UNSURE
+    else:
+        endness = Endness.UNSURE
+    ident = ident.lower()
+    cls = None
+    aendness = None
+    for arxs, abits, aendness, acls in arch_id_map:
+        found_it = False
+        for rx in arxs:
+            if re.search(rx, ident):
+                found_it = True
+                break
+        if not found_it:
+            continue
+        if bits and bits != abits:
+            continue
+        if aendness == Endness.ANY or endness == aendness or endness == Endness.UNSURE:
+            cls = acls
+            break
+    if not cls:
+        raise ArchNotFound(
+            f"Can't find architecture info for architecture {ident} with {repr(bits)} bits and {endness} endness"
+        )
+    if endness == Endness.UNSURE:
+        if aendness == Endness.ANY:
+            # We really don't care, use default
+            return cls(cls.default_endness)
+        else:
+            # We're expecting the ident to pick the endness.
+            # ex. 'armeb' means obviously this is Iend_BE
+            return cls(aendness)
+    else:
+        return cls(endness)
+
+
+def get_host_arch():
+    """
+    Return the arch of the machine we are currently running on.
+    """
+    return arch_from_id(_platform.machine())
+
 
 __all__ = [
     "Arch",
@@ -65,22 +150,3 @@ __all__ = [
     "register_arch",
     "reverse_ends",
 ]
-
-
-def arch_from_id(ident: str, endness: str = Endness.ANY, bits: str | int = "") -> Arch:
-    """
-    Take our best guess at the arch referred to by the given identifier, and return an instance of its class.
-
-    You may optionally provide the ``endness`` and ``bits`` parameters to help this function out. ``bits`` is
-    either a number of bits or a string containing one, which is what an ELF class is.
-
-    A full sleigh language id, such as ``pa-risc:BE:32:default``, returns the ArchPcode for that language. It
-    carries its own endness and width, so the ``endness`` and ``bits`` hints do not apply to it.
-    """
-    # A language id names one language, so it answers before the registered architectures, whose regexes
-    # would otherwise claim ARM:LE:32:v7 for ArchARMEL.
-    if ":" in ident:
-        with contextlib.suppress(ArchError):
-            return ArchPcode(ident)
-
-    return _arch_from_id(ident, endness, bits)
--- a/tests/test_pcode.py
+++ b/tests/test_pcode.py
@@ -95,14 +95,6 @@ class TestArchPcode(unittest.TestCase):
         with self.assertRaises(ArchNotFound):
             arch_from_id("EM_PARISC", "be", 32)
 
-    def test_arch_from_id_without_pypcode(self):
-        # archinfo installed without the pcode extra has no language to reach, and answers exactly
-        # as it did before language ids were accepted at all.
-        with patch("archinfo.arch_pcode._has_pypcode", False):
-            assert arch_from_id("x86").name == "X86"
-            with self.assertRaises(ArchNotFound):
-                arch_from_id("pa-risc:BE:32:default")
-
     def test_pickle(self):
         arch = ArchPcode("68000:BE:32:default")
         pickle.dumps(arch)
@@ -200,6 +192,18 @@ class TestArchPcode(unittest.TestCase):
         with self.assertRaises(ArchError) as cm:
             s390x.pcode_arch()
         assert "does not have a pcode_arch defined" in str(cm.exception)
+
+
+class TestArchFromIdWithoutPypcode(unittest.TestCase):
+    def test_arch_from_id_without_pypcode(self):
+        # archinfo installed without the pcode extra has no language to reach, and answers exactly as it
+        # did before language ids were accepted at all. Each guard is exercised on its own: archinfo.arch
+        # skips the branch when pypcode is absent, and ArchPcode refuses to construct without it.
+        for target, absent in (("archinfo.arch._pypcode", None), ("archinfo.arch_pcode._has_pypcode", False)):
+            with patch(target, absent):
+                assert arch_from_id("x86").name == "X86"
+                with self.assertRaises(ArchNotFound):
+                    arch_from_id("pa-risc:BE:32:default")
 
 
 if __name__ == "__main__":
--- a/tests/test_pcode.py
+++ b/tests/test_pcode.py
@@ -61,8 +61,8 @@ class TestArchPcode(unittest.TestCase):
         assert arch_from_id("Xtensa:BE:32:default", "le", 64).pcode_id == "Xtensa:BE:32:default"
 
     def test_arch_from_id_by_language_id_of_a_registered_arch(self):
-        # 74 of the 187 languages pypcode ships name a processor arch_id_map also matches, so an id
-        # has to answer before that map or naming one of them silently returns the VEX architecture.
+        # arch_id_map resolves 74 of the 187 language ids pypcode ships to a VEX architecture, so an id
+        # has to answer before that map or naming one of them silently returns that architecture.
         for language_id in ["ARM:LE:32:v7", "MIPS:BE:32:default", "PowerPC:BE:64:default", "x86:LE:16:Real Mode"]:
             arch = arch_from_id(language_id)
             assert isinstance(arch, ArchPcode)
@@ -196,14 +196,12 @@ class TestArchPcode(unittest.TestCase):
 
 class TestArchFromIdWithoutPypcode(unittest.TestCase):
     def test_arch_from_id_without_pypcode(self):
-        # archinfo installed without the pcode extra has no language to reach, and answers exactly as it
-        # did before language ids were accepted at all. Each guard is exercised on its own: archinfo.arch
-        # skips the branch when pypcode is absent, and ArchPcode refuses to construct without it.
-        for target, absent in (("archinfo.arch._pypcode", None), ("archinfo.arch_pcode._has_pypcode", False)):
-            with patch(target, absent):
-                assert arch_from_id("x86").name == "X86"
-                with self.assertRaises(ArchNotFound):
-                    arch_from_id("pa-risc:BE:32:default")
+        # archinfo installed without the pcode extra has no language to reach, so ArchPcode refuses to
+        # construct and arch_from_id answers exactly as it did before language ids were accepted at all.
+        with patch("archinfo.arch_pcode._has_pypcode", False):
+            assert arch_from_id("x86").name == "X86"
+            with self.assertRaises(ArchNotFound):
+                arch_from_id("pa-risc:BE:32:default")
 
 
 if __name__ == "__main__":
--- a/tests/test_pcode.py
+++ b/tests/test_pcode.py
@@ -1,8 +1,9 @@
 # pylint:disable=missing-class-docstring,no-self-use
 import pickle
 import unittest
+from unittest.mock import patch
 
-from archinfo import ArchError, ArchPcode, ArchS390X, Endness, arch_from_id
+from archinfo import ArchError, ArchNotFound, ArchPcode, ArchS390X, Endness, arch_from_id
 
 try:
     import pypcode
@@ -46,6 +47,61 @@ class TestArchPcode(unittest.TestCase):
     def test_arch_bad_langid(self):
         with self.assertRaises(ArchError):
             ArchPcode("invalid")
+
+    def test_arch_from_id_by_language_id(self):
+        # A full sleigh language id names one language, and nothing registered can name it at all:
+        # pypcode is the only definition of PA-RISC that archinfo has.
+        arch = arch_from_id("pa-risc:BE:32:default")
+        assert isinstance(arch, ArchPcode)
+        assert arch.pcode_id == "pa-risc:BE:32:default"
+        assert arch.bits == 32
+        assert arch.memory_endness == Endness.BE
+
+        # The id carries its own endness and width, so a hint that contradicts it does not apply.
+        assert arch_from_id("Xtensa:BE:32:default", "le", 64).pcode_id == "Xtensa:BE:32:default"
+
+    def test_arch_from_id_by_language_id_of_a_registered_arch(self):
+        # 74 of the 187 languages pypcode ships name a processor arch_id_map also matches, so an id
+        # has to answer before that map or naming one of them silently returns the VEX architecture.
+        for language_id in ["ARM:LE:32:v7", "MIPS:BE:32:default", "PowerPC:BE:64:default", "x86:LE:16:Real Mode"]:
+            arch = arch_from_id(language_id)
+            assert isinstance(arch, ArchPcode)
+            assert arch.pcode_id == language_id
+
+    def test_arch_from_id_round_trips_every_language(self):
+        # angr's database serializer stores an architecture as (name, memory_endness, bits) and
+        # reloads it with arch_from_id, and ArchPcode.name is the language id.
+        assert pypcode is not None
+        for pcode_arch in pypcode.Arch.enumerate():
+            for language in pcode_arch.languages:
+                arch = ArchPcode(language)
+                restored = arch_from_id(arch.name, arch.memory_endness, arch.bits)
+                assert isinstance(restored, ArchPcode)
+                assert restored.pcode_id == language.id
+
+    def test_arch_from_id_prefers_registered_arches(self):
+        # An identifier that is not a language id resolves exactly as it always did.
+        for ident in ["x86", "amd64", "arm", "aarch64", "mips32", "mips64", "ppc32", "ppc64", "s390x", "riscv64"]:
+            assert not isinstance(arch_from_id(ident), ArchPcode)
+
+    def test_arch_from_id_without_a_language(self):
+        # An identifier shaped like a language id that names no language is still not found.
+        with self.assertRaises(ArchNotFound):
+            arch_from_id("nosucharch:LE:32:default")
+        # pypcode has no little-endian SPARC variant of this language.
+        with self.assertRaises(ArchNotFound):
+            arch_from_id("sparc:LE:32:default")
+        # A machine name is not a language id, and is not searched for as one.
+        with self.assertRaises(ArchNotFound):
+            arch_from_id("EM_PARISC", "be", 32)
+
+    def test_arch_from_id_without_pypcode(self):
+        # archinfo installed without the pcode extra has no language to reach, and answers exactly
+        # as it did before language ids were accepted at all.
+        with patch("archinfo.arch_pcode._has_pypcode", False):
+            assert arch_from_id("x86").name == "X86"
+            with self.assertRaises(ArchNotFound):
+                arch_from_id("pa-risc:BE:32:default")
 
     def test_pickle(self):
         arch = ArchPcode("68000:BE:32:default")
--- a/tests/test_register_arch.py
+++ b/tests/test_register_arch.py
@@ -1,0 +1,41 @@
+# pylint:disable=missing-class-docstring,no-self-use
+import re
+import unittest
+
+from archinfo import ArchX86, Endness, all_arches, arch_from_id, register_arch
+from archinfo.arch import _all_arches_set, arch_id_map
+
+
+class ArchThrowaway(ArchX86):
+    def __init__(self, endness=Endness.LE):
+        super().__init__(endness)
+        self.name = "THROWAWAY"
+
+
+class TestRegisterArch(unittest.TestCase):
+    def setUp(self):
+        self._id_map = list(arch_id_map)
+        self._all_arches = list(all_arches)
+        self._arch_set = set(_all_arches_set)
+
+    def tearDown(self):
+        arch_id_map[:] = self._id_map
+        all_arches[:] = self._all_arches
+        _all_arches_set.clear()
+        _all_arches_set.update(self._arch_set)
+
+    def test_register_arch_accepts_a_compiled_regex(self):
+        # register_arch documents "str or compiled regular expression", and the check for the second
+        # kind read re._pattern_type, which Python dropped in 3.7, so a compiled pattern raised
+        # AttributeError instead of registering.
+        register_arch([re.compile("^throwaway$")], 32, Endness.LE, ArchThrowaway)
+        assert isinstance(arch_from_id("throwaway"), ArchThrowaway)
+
+    def test_register_arch_still_rejects_other_types(self):
+        not_a_regex: list = [42]
+        with self.assertRaises(TypeError):
+            register_arch(not_a_regex, 32, Endness.LE, ArchThrowaway)
+
+
+if __name__ == "__main__":
+    unittest.main()


zardus:

Validation record for head `c24c173cfa01d42cf22dd207f997f00d8dbe57ec` against baseline `384c676de767e1216a7fee2a08f2d549ff2a1028`, so everything below is this PR's own. The sections are dated and later ones win: the 2026-09-05 section at the bottom lists the figures above it that the current head has moved. The figures were measured at `20d78c28173006eda13b6b60cb4fc901b5fdfda5` against `160613e2fbd62c72d9e1d4046b5d698fa9493b43`, this branch's own copy of #363's commit; the currency checks at the bottom carry them to this head. Python 3.12.13, pypcode 4.0.1.dev0, pyelftools 0.33, pefile 2024.8.26.

- Regression: `python -m pytest tests/test_pcode.py` — 15 tests; on the baseline `test_arch_from_id_by_language_id`, `test_arch_from_id_by_language_id_of_a_registered_arch` and `test_arch_from_id_round_trips_every_language` fail, all 15 pass on head
- Focused: `python -m pytest tests` — 26 passed, 0 skipped, 0 xfailed
- Complete workspace gate on this head, with `angr` adopted so its suite is not skipped: archinfo 26 passed; angr Python 2471 passed, 46 skipped, 2 xfailed, 260 subtests passed; angr Rust 35 passed, 0 failed; plus environment, workspace checks, test-inputs, all pre-commit hooks and per-feature instances. `cle`, `pyvex`, `claripy`, `pypcode` and `angr-management` were skipped as unadopted; none is a caller this change reaches, and angr is, which is why it was adopted rather than left out.
- Lint/type, per changed file against the merge base `d17a90a`: pylint `archinfo/arch.py` 9.70 -> 9.70, `archinfo/arch_pcode.py` 10.00 -> 10.00, `tests/test_pcode.py` 10.00 -> 10.00; pyright badness `archinfo/arch.py` 0.2227 -> 0.2199, `archinfo/arch_pcode.py` 0.0 -> 0.0, `tests/test_pcode.py` 0.0 -> 0.0. `arch_pcode.py` is in the changed set only because of this branch's #363 commit; this PR no longer touches it
- Hooks: `pre-commit run` over the changed files — passed, tree unchanged
- Negative configuration: with `archinfo.arch_pcode._has_pypcode` set to False, standing in for archinfo installed without the `pcode` extra, `arch_from_id` answers exactly as it does on the baseline. Pinned by `test_arch_from_id_without_pypcode`

The round trip angr's database serializer performs, run through `LoadArgsJSONEncoder` and `LoadArgsJSONDecoder` from `angr/angrdb/serializers/loader.py` rather than simulated:

| Stored `arch.name` | Baseline | Head |
| --- | --- | --- |
| `pa-risc:BE:32:default` | `ArchNotFound` | `ArchPcode` `pa-risc:BE:32:default` |
| `x86:LE:16:Real Mode` | `ArchNotFound` | `ArchPcode` `x86:LE:16:Real Mode` |
| `ARM:LE:32:v7` | `ArchARMEL` | `ArchPcode` `ARM:LE:32:v7` |
| `MIPS:BE:32:default` | `ArchMIPS32` | `ArchPcode` `MIPS:BE:32:default` |

The second failure mode is the quiet one: 70 of the 187 ids come back as the VEX architecture of the same processor when they are reloaded with their stored endness and bits, so reloading such a project silently swapped the engine. All 187 round-trip on head; `test_arch_from_id_round_trips_every_language` pins it.

Resolver comparison, baseline against head, over 492 identifiers x 7 endness/width hint combinations = 3444 resolutions. The identifiers are every `all_arches` name, every pyelftools `ENUM_E_MACHINE` name, every `pefile` `IMAGE_FILE_MACHINE_*` name, every sleigh language id and processor name, and the literal strings cle's ELF, PE, COFF, TE and Mach-O backends pass.

- Unchanged answers: 2135
- `ArchNotFound` -> resolved: 857; answer changed: 452; resolved -> `ArchNotFound`: 0; new exception types: 0
- Every identifier whose answer differs at any hint is one of the 187 language ids, and every changed answer is exactly the language named. None of the 236 identifiers cle's backends can pass is a language id, so no loader's answer changes.
- Cost of a miss on the path cle's ELF backend takes for a machine nothing registered claims, `arch_from_id("EM_LOONGARCH", "le", 64)`, mean of 200 calls measured back to back: 0.011 ms on baseline and on head. The processor-name search this replaces cost 12.7 ms there.

Why a name search was dropped rather than kept, measured over every `ENUM_E_MACHINE` value at both ELF classes and both byte orders (752 combinations), comparing it against `ELF._get_compatible_pcode_languages`, which cle already consults when `arch_from_id` raises:

- Combinations only the name search resolved: 3 — `EM_PARISC` big-endian 32, `EM_SPARC` big-endian 64, and `EM_BPF` at 32-bit class, which no real eBPF object uses
- Combinations where it preempted an opinion cle would have used: 13, of which 6 got a different answer, including LoongArch, where the opinion is decided by `e_flags` that archinfo cannot see
- Combinations the opinion path resolves and the name search did not: 23, including `EM_68K`, `EM_SH`, `EM_MSP430`, `EM_AVR` and `EM_SPARCV9`

Caveats:

- A language id's own endness and width win over the `endness` and `bits` hints, which the id already carries. The match is exact: `arch_from_id` hands the id to `ArchPcode`, whose constructor compares it against `language.id` case-sensitively.
- Nothing that reads an architecture out of a file header passes a language id, so this changes no loader's answer and is invisible to a corpus sweep. Its reach is a stored architecture read back and an explicit `arch=`.
- Retarget sequence once #363 merges: rebase `feature/fix-archinfo-arch-coverage` onto `origin/master`, which drops the duplicated commit, rerun the focused suite and the lint and type comparison on the rebased head, then mark ready.

---

## Superseded, 2026-08-18: everything this comment previously recorded

The earlier record was taken on `08c021f`, which also resolved a machine name such as `EM_PARISC` by searching sleigh processor names with normalization and ranking. Head `9c6ba26` matches full language ids only, so the claims that rested on that search are void:

- The end-to-end PA-RISC load of `tests/hppa/test-instr_hppa` no longer comes from here. That file's `e_flags` is `0x210`, exactly the `secondary` constraint on Ghidra's PA-RISC ELF opinion, so the loader can decide it from the header: cle master's `_get_compatible_pcode_languages` returns nothing for that file and angr/cle#740's returns `pa-risc:BE:32:default`. PA-RISC loading belongs to that PR, at the layer that can read `e_flags`.
- The 2026-08-15 corpus measurement of the `ArchNotFound` class, including the SPARC blob that segfaulted inside pypcode, described objects `08c021f` reached from machine names. `9c6ba26` reaches none of them, so its corpus reach is zero. The segfault is real and independent of this PR.
- The earlier resolver comparison, 500 identifiers and 1069 newly resolved, is replaced by the 492-identifier comparison above.

---

## Superseded, 2026-08-19: the classmethod and the `arch_pcode.py` changes

Head `9c6ba26` reached the languages through a new `ArchPcode.from_id` classmethod and rewrote `arch_pcode.py`'s pypcode import guard. Head `20d78c2` calls `ArchPcode(ident)` directly and leaves `arch_pcode.py` at master's content, so:

- The `archinfo/arch_pcode.py` pyright badness row above reads 0.0 -> 0.0 rather than 0.2381 -> 0.1579; the file is unchanged by this PR.
- Matching is now case-sensitive, because `ArchPcode._get_language_by_id` compares ids exactly where `from_id` lowercased both sides. `arch_from_id("PA-RISC:be:32:DEFAULT")` raises `ArchNotFound` on this head and the assertion that covered it is gone.
- Every other answer is unchanged: all 3444 resolutions in the comparison above are identical on `9c6ba26` and `20d78c2`, and the angrdb round-trip table was re-run on this head.

## Head currency, checked 2026-08-27 — head `cfa6d72ed75393df79dbb4dfc7486e2aeeb4b41b`

Everything above was measured at `20d78c28173006eda13b6b60cb4fc901b5fdfda5`, and still describes this head:

- `git range-diff 20d78c2~2..20d78c2 cfa6d72~2..cfa6d72` reports both commits unchanged (`160613e = 73a5917`, `20d78c2 = cfa6d72`), so the rebase moved only the base.
- `python -m pytest tests` re-run at this head — 26 passed, matching the row above.
- Corpus: the `dec-snapshots` branch for this pull request rests on `dec-snapshots` master `0373819651d8f091d543e753c5d814bbfc98ba45` with `Decompiler Snapshot Testing` green, so the output is byte-identical. That is silence rather than coverage: the new branch is gated on `":" in ident`, and nothing that loads an ELF passes a sleigh language id.

~~Still a draft, and still parked on #363 for the reason above: this branch carries #363's commit because accepting any language id means all 187 have to construct, and archinfo squash-merges, so landing this one first would fold that commit in and leave #363 with a diff that no longer applies. Nothing references this pull request, so no other branch is being tested against it in the meantime. Once #363 merges: rebase, which drops the duplicate, re-run the focused suite and the lint and type comparison, then mark ready.~~

## Rebased and taken out of draft, 2026-08-28 — head `d4e902b96368accacaad88c51b0dfb63e1cfa734`

#363 merged on 2026-08-27 as `f92307b32e138de45b6ed4e8ca7794b67fdff6f8`, which is the only thing this draft was parked on. The retarget sequence recorded above was run exactly as written:

- `git rebase --onto origin/master 73a5917 cfa6d72`. `git range-diff bf85c7e..cfa6d72 f92307b..d4e902b` reports `73a5917 < -` — the duplicated #363 commit dropped, as intended — and `cfa6d72 = d4e902b`, this PR's own commit unchanged. No conflict; `gh pr view` reported `MERGEABLE`/`CLEAN` before and after
- The diff against master is now `archinfo/arch.py` +13 and `tests/test_pcode.py` +57. `archinfo/arch_pcode.py` is not touched at all, which is what the 2026-08-19 reply promised and what the duplicate commit was hiding
- `python -m pytest -q tests` re-run on this head — 26 passed, matching the focused row above
- `pre-commit run --files archinfo/arch.py tests/test_pcode.py` — passed, tree unchanged
- Both CHANGES_REQUESTED items were answered on `20d78c2` and remain answered here, since the commit is unchanged: `ArchPcode.from_id` is gone and `arch_from_id` calls the constructor under `contextlib.suppress(ArchError)`; the guard is folded into the single `if ":" in ident:` and the comment is two lines
- The ordering concern raised against angr/cle#740 no longer applies: matching is exact language ids only, so nothing a loader passes reaches this branch and `ELF._get_compatible_pcode_languages` is never preempted

Out of draft on this evidence.

## Review fix, 2026-08-28 — head `3c2f1f6884d2e01f8d445849e99819d188bd2090`

Rebased onto `b8ffe85`, which is master with #371 merged; that commit rewrites `reverse_ends` and touches `arch.py` outside `arch_from_id`. `git range-diff f92307b..d4e902b b8ffe85..9138acf` reports `d4e902b = 9138acf`, this PR's own commit unchanged. `3c2f1f6` then answers the review comment on `arch.py:879`: the language-id branch is guarded on a module-level optional `pypcode` import, in the shape `arch.py` already uses for `_pyvex`, `_unicorn`, `_capstone` and `_keystone`, and the without-pypcode regression moves out of the `skipUnless(pypcode is not None)` class so it runs in both configurations. Superseding the rows above:

- Focused: `python -m pytest tests` with pypcode installed — 30 passed, 34 subtests passed, identically on `9138acf` and `3c2f1f6`. Replaces the "26 passed" row; the four extra tests and the subtests come from #371, not from this branch
- Negative configuration, now measured rather than mocked: the suite run under a `sys.meta_path` finder whose `find_spec` raises `ModuleNotFoundError` for `pypcode`, with the block asserted to fire before pytest starts. `9138acf` — 15 passed, 15 skipped, `test_arch_from_id_without_pypcode` **skipped**; `3c2f1f6` — 16 passed, 14 skipped, that test **passed**. Replaces the mock-only row above, which could not run in the configuration it described. `find_spec` is the hook that decides this: a blocker written against the legacy `find_module` is ignored on Python 3.12 and blocks nothing
- Lint/type per changed file against the merge base `b8ffe85`: pylint `archinfo/arch.py` 9.70 -> 9.70, `tests/test_pcode.py` 10.00 -> 10.00; pyright badness `archinfo/arch.py` 0.22175 -> 0.21784, `tests/test_pcode.py` 0.0 -> 0.0. Replaces the row keyed to merge base `d17a90a`
- Output comment: both captures regenerated at `b8ffe85` and `3c2f1f6` and byte-identical to the published text, so nothing a user who has the `pcode` extra sees has moved. Its before-side provenance line was re-keyed from `f92307b` to `b8ffe85`
- Hooks: every configured `pre-commit` hook ran on the commit — passed, tree unchanged
- Consumers: with this branch's checkout first on `PYTHONPATH`, `pyvex`, `cle` and `angr` import and `angr.Project("/bin/true").arch` is `<Arch AMD64 (LE)>`; the ten registered identifiers and the language ids resolve as tabulated above

Not run at this head: the complete workspace gate. Its environment step fails at `python -m pip check` for a reason outside this branch — `angr-megabench 0.1.0 has requirement pyxbe==1.0.3, but you have pyxbe 1.0.4` — and `repos/angr-management` is 9 commits behind its `origin/master`. The archinfo suite, the pre-commit hooks and the merge-base lint and type comparison were run in a detached worktree of this branch instead, and are the rows above. `angr`'s suite was last run against this branch at `20d78c2`; the only behaviour `3c2f1f6` changes is on a machine where pypcode is absent, since `_pypcode is not None` is true wherever the extra is installed and the branch condition is then exactly what it was.

## Second review comment, 2026-08-28 — head `dd03d23356d6c622b07bfc30fad9ae6fd183e215`

`3c2f1f6` reached terminal CI before this push: 19 checks passing, 0 failing, read twice — `gh pr checks` and the `statusCheckRollup` query that accepts both `conclusion` and `state` — at 16:47Z.

`dd03d23` answers the second comment on `arch.py:884` (`"Remove the local import since there is a top-level one"`) and is comment-only: 2 insertions, 1 deletion in `archinfo/arch.py`, no statement changed. The top-level import that prompted it is at line 15 inside `if TYPE_CHECKING:`, so it has no runtime binding. The three ways of removing the runtime import were measured rather than argued:

- Delete it and rely on the `TYPE_CHECKING` import — `import archinfo` succeeds and `arch_from_id("ARM:LE:32:v7")` raises `NameError: name 'ArchPcode' is not defined`
- Promote the `TYPE_CHECKING` import to module scope — the package stops importing: `ImportError: cannot import name 'Arch' from partially initialized module 'archinfo.arch' (most likely due to a circular import)`, because `arch_pcode` does `from .arch import Arch, Endness, Register` before `Arch` exists
- A module-scope import at the bottom of `arch.py`, below the class definitions — this works, and pylint stays at 9.70; it costs `# noqa: E402` and `# pylint: disable=wrong-import-position`, and makes `Arch.pcode_arch`'s own deferred import redundant, which is master's line rather than this PR's

The import therefore stays, matching `Arch.pcode_arch` on master, whose comment reads `# Delayed import to avoid circular dependency`. `arch.py` names `ArchPcode` in three places — the `TYPE_CHECKING` import, `pcode_arch`'s runtime import, and `arch_from_id`'s — and only the last belongs to this PR. `dd03d23` says all of that in the code so the next reader does not have to ask.

- Focused, unchanged at this head: `python -m pytest tests` with pypcode installed — 30 passed, 34 subtests passed
- Negative configuration, unchanged at this head: under the `find_spec` blocker — 16 passed, 14 skipped, `test_arch_from_id_without_pypcode` passed, with the block asserted to fire before pytest started
- Lint/type per changed file against the merge base `b8ffe85`: pylint `archinfo/arch.py` 9.70 -> 9.70, `tests/test_pcode.py` 10.00 -> 10.00; pyright badness `archinfo/arch.py` 0.22175 -> 0.21762, `tests/test_pcode.py` 0.0 -> 0.0
- Output comment: the capture regenerated at this head is byte-identical to the one published, as a comment-only commit requires; its after-side provenance line was re-keyed from `3c2f1f6` to `dd03d23`
- Method note, because it nearly cost the measurement: the two runtime `from .arch_pcode import ArchPcode` lines in `arch.py` are byte-identical including indentation and the trailing pylint pragma, so a variant built by string replacement edits both. Each variant above asserted the occurrence count before and after its edit, and the first attempt aborted on that assertion — without it the three variants would have been unmodified copies of the head, and all three would have "passed"

### Terminal state for this head, checked 2026-08-28T17:24Z

- CI: 19 checks passing, 0 failing on `dd03d23356d6c622b07bfc30fad9ae6fd183e215`, read twice — `gh pr checks` and the `statusCheckRollup` query that accepts both `conclusion` and `state` — and `MERGEABLE`. The run scheduled 16 check runs under one workflow run, so this is a real pass rather than a missed schedule
- Corpus: `ci / Decompiler Snapshot Testing` succeeded at 16:58:09Z and the `angr/archinfo_365` branch rests exactly on `dec-snapshots` master `f79001702a2a32bab64ae64412f763e48afb92e5`, ahead 0 and behind 0, so the corpus output is byte-identical at this head. That is silence rather than coverage: the corpus is x86-64 CGC decompiled through VEX, and the new branch is gated on `":" in ident`, which nothing that loads an ELF passes

### Re-keyed 2026-08-29 — head `14c81fc693879ebfd027d2e648269c276360d269`

Re-keyed from `dd03d23356d6c622b07bfc30fad9ae6fd183e215` after a rebase onto
`7e72c6caed422219d97af234f7acec96b37535c4`, which the branch needed once #375 merged.
`git range-diff b8ffe857..dd03d23 7e72c6ca..14c81fc` marks all three commits `!` rather
than `=`, and the whole of that is one line: master now carries the `bits: str | int`
annotation on `arch_from_id` that this branch used to add, so the branch's signature hunk
adds only `endness: str`. Everything else in the branch's diff against its merge base is
byte-identical — the optional `pypcode` import, the `":" in ident` early return through
`ArchPcode`, the docstring paragraph, the `TYPE_CHECKING` comment, and all 62 lines of
`tests/test_pcode.py`. The behaviour under test did not move, so every figure above still
describes this patch, and so does the corpus measurement in
https://github.com/angr/archinfo/pull/365#issuecomment-5460633871 — 951 of the 1,212
`ArchNotFound` failures in that sample — which was taken at `dd03d23` on exactly this
`arch_from_id` body.

- CI at this head is green and terminal, read 2026-08-29T21:00Z: 17 check runs all
  `success` under run [33270369274](https://github.com/angr/archinfo/actions/runs/33270369274),
  plus `pre-commit.ci - pr` and `docs/readthedocs.org:archinfo` — 19 in all, with the
  combined status `success` and no queued or failing leg.

## Moved into `__init__.py`, 2026-08-30 — head `26998a916dc1526897432e650777b83368b2672c`

twizmwazin asked for the resolution to live in `__init__.py`
(https://github.com/angr/archinfo/pull/365#issuecomment-5465187586). This head puts it
there and `arch.py` loses all four pieces it needed to host it: the `contextlib` import,
the module-level `pypcode` sentinel, the function-local `ArchPcode` import with its pylint
suppression, and the comment explaining the delay. `arch.py`'s diff against the merge base
is now the `endness: str` annotation and nothing else. The sentinel was redundant:
`ArchPcode.__init__` already raises `ArchError("pypcode not installed")`, which the
surrounding `contextlib.suppress(ArchError)` catches, so one branch covers a missing
dependency and an unknown language. This settles the `arch.py:884` comment the other way
from the 2026-08-28 section above: the runtime `ArchPcode` import is not kept after all,
because `arch_from_id` is no longer the thing that needs the name.

Three counts describe the 187 languages at the merge base and they are not
interchangeable, which is how a wrong figure reached this record and the code. Re-derived
here against master's `arch_from_id`, pypcode 4.0.1.dev0:

| Quantity | Count | Raising or not matching |
| --- | ---: | ---: |
| ids matching some `arch_id_map` regex | 78 | 109 |
| ids resolving through a bare `arch_from_id(id)`, no hints | 74 | 113 |
| ids returning a VEX architecture from `arch_from_id(name, memory_endness, bits)` | 70 | 117 |

They nest, 78 > 74 > 70. The four dropped between the first and the second
(`RISCV:LE:32:AndeStar_v5`, `RISCV:LE:32:default`, `x86:LE:64:compat32`,
`x86:LE:64:default`) match a regex and then fail its bits or endness filter; the four
dropped between the second and the third (`AARCH64:BE:32:ilp32`, `AARCH64:LE:32:ilp32`,
`x86:LE:16:Protected Mode`, `x86:LE:16:Real Mode`) resolve bare and stop resolving once
the stored endness and bits are supplied. The quiet-failure sentence above named the
matching count where it meant the round-trip one and now reads 70; the comment in
`test_arch_from_id_by_language_id_of_a_registered_arch` describes the bare call that
test makes and now reads 74.

Every figure above still describes this head, measured rather than argued.
`arch_from_id` was dumped over 548 identifiers x 7 endness/width hint combinations = 3836
resolutions on `14c81fc` and on this head, and the two dumps are byte-identical. The
identifier set is every `all_arches` name, every pyelftools `ENUM_E_MACHINE` name, every
pefile `IMAGE_FILE_MACHINE_*` name, every sleigh language id and processor name, and the
literals cle's ELF, PE, COFF, TE and Mach-O backends pass. Against the merge base the same
dump gives 2527 unchanged, 833 raise -> resolved, 476 answer changed and 0 resolved ->
raise; the 187 identifiers that differ at any hint are exactly the 187 sleigh language ids,
and every changed answer is the language named. The corpus rates in
https://github.com/angr/archinfo/pull/365#issuecomment-5460633871 carry for the same
reason: with pypcode installed, the old `_pypcode is not None and ":" in ident` and the new
`":" in ident` are the same condition, and that comment's figures were taken with pypcode
installed.

- Regression: `python -m pytest tests -q` with pypcode installed — 37 passed, 34 subtests
- Negative configuration, measured rather than mocked: the same suite in a virtualenv with
  neither pypcode nor pyvex installed — 20 passed, 17 skipped, 34 subtests, with
  `import pypcode` asserted to fail first. There `import archinfo` works,
  `arch_from_id("x86")` returns `ArchX86`, `arch_from_id("ARM:LE:32:v7")` returns
  `ArchARMEL` and `arch_from_id("pa-risc:BE:32:default")` raises `ArchNotFound` — the merge
  base's answers exactly
- No import cycle: importing `archinfo.arch`, `archinfo.arch_pcode` or `archinfo` first all
  succeed and all reach the same `arch_from_id`, in both configurations
- Lint/type per changed file against the merge base `7e72c6c`, now three files: pylint
  `archinfo/__init__.py` 10.00 -> 10.00, `archinfo/arch.py` 9.70 -> 9.70,
  `tests/test_pcode.py` 10.00 -> 10.00; pyright badness `archinfo/__init__.py` 0.0 -> 0.0,
  `archinfo/arch.py` 0.22152 -> 0.22152, `tests/test_pcode.py` 0.0 -> 0.0
- Hooks: every configured `pre-commit` hook over all files — passed, tree unchanged
- Docs: `sphinx-build -b html docs` at the merge base, at `14c81fc` and at this head emits
  the same 12 warnings, character for character
- Downstream, with this branch first on `PYTHONPATH`: `cle.Loader` reads AMD64, ARMEL and
  X86 out of `binaries/tests/{x86_64,armel,i386}/fauxware`, `angr.Project` infers AMD64 and
  accepts `arch="AMD64"`, and `LoadArgsJSONEncoder`/`LoadArgsJSONDecoder` round-trip all
  four language ids in the table above. No consumer reaches past the package: every
  `arch_from_id` caller in angr, cle, pyvex and angr-management writes
  `archinfo.arch_from_id`, and none imports it from `archinfo.arch`
- Cost of a miss on the path cle's ELF backend takes, `arch_from_id("EM_LOONGARCH", "le",
  64)`, min of 5 batches of 2000 calls: 0.0104 ms at the merge base, 0.0104 ms at
  `14c81fc`, 0.0084 ms here. The host is running a corpus sweep and the ordering flips
  between runs, so this says only that the added frame costs nothing measurable

Not run at this head: the complete workspace gate, and CI, which has not seen this head
yet. A corpus sweep holds the shared virtualenv and `run-all-tests.sh` enters `nix
develop`, which would reinstall the editable packages and rewrite native libraries that
sweep has mapped, so archinfo was validated alone in a detached worktree with its own
virtualenv, in both configurations above. The angr, cle, claripy, pyvex and
angr-management suites and angr's Rust tests did not run. What stands in for them is the
3836-resolution comparison above: this head answers identically to `14c81fc`, which is the
head 17 green check runs were read on, and the only import this commit adds anywhere is
`contextlib` in `__init__.py`, which `arch.py` used to carry.

## Whole function moved into `__init__.py`, rebased, 2026-09-05 — head `c24c173cfa01d42cf22dd207f997f00d8dbe57ec`

twizmwazin asked for the whole function rather than a wrapper, and above `__all__`
(https://github.com/angr/archinfo/pull/365#issuecomment-5480854320). `arch_from_id` now
lives in `__init__.py` in full, `get_host_arch` moved with it, and `arch.py` carries
neither. The branch is also rebased onto master `384c676` (9.3.5.dev0); `git range-diff`
reports the four earlier commits replayed unchanged, the only difference being the
`__version__` context line master bumped.

What the 2026-08-30 section above no longer describes, and what replaces it:

- "`arch.py`'s diff against the merge base is now the `endness: str` annotation and
  nothing else" — false at this head. That annotation went away with the function; the
  file's diff is the deletion of `arch_from_id` and `get_host_arch`, its now-unused
  `platform` import, and the `re.Pattern` repair below.
- "548 identifiers x 7 ... = 3836 resolutions" — re-run at this head over 438 identifiers
  x 7 hint combinations = 3066 resolutions, the pool being every `all_arches` name, every
  pyelftools `ENUM_E_MACHINE` name, every pefile `IMAGE_FILE_MACHINE_*` name, all 187
  sleigh language ids, and 22 literal identifiers of which 14 are not already in those
  sets. Byte-identical to the same dump on
  `26998a9` replayed on this base. Against master `384c676` it differs in 1309 rows, every
  one of them a language id, and 0 go from resolved to raising.
- "37 passed, 34 subtests" — now 46 passed, 34 subtests. Seven of the nine came from
  master between the two bases, not from this branch: `pytest tests --collect-only` on an
  archive of each revision gives 31 at `7e72c6c`, 38 at `384c676`, 44 with this branch's
  `test_pcode.py` cases, and 46 with `tests/test_register_arch.py`, which adds two.
- "20 passed, 17 skipped" for the negative configuration — re-run here as a `sys.meta_path`
  finder whose `find_spec` raises for `pypcode`, asserted to fire before the check: `import
  archinfo` works, `arch_from_id("x86")` is `ArchX86`, `get_host_arch()` answers, and
  `arch_from_id("pa-risc:BE:32:default")` raises `ArchNotFound` — master's answers exactly,
  and identical on `26998a9` replayed and on this head.
- "Lint/type per changed file against the merge base `7e72c6c`, now three files ... pyright
  badness" — the merge base is `384c676`, the changed set is four files, and ci-settings
  retired the per-line badness score on 2026-09-01 (`b23d782`), so the Typecheck job counts
  errors. At this head: pylint `archinfo/__init__.py` 10.00 -> 10.00, `archinfo/arch.py`
  9.70 -> 9.77, `tests/test_pcode.py` 10.00 -> 10.00, `tests/test_register_arch.py` new at
  10.00; pyright errors 0 -> 0, 21 -> 20, 0 -> 0, and 0 for the new file.

New at this head:

- `register_arch`'s type check read `re._pattern_type`, removed from Python in 3.7. A list
  of plain strings short-circuits before reaching it, which is why archinfo's own 21
  registrations work; any list holding something else raised `AttributeError`. Fixed to
  `re.Pattern`, with the annotation and the docstring brought into line, and covered by
  `tests/test_register_arch.py`, whose two cases both fail on master with that
  `AttributeError` at `arch.py:841`. It is in this branch because `arch.py` would otherwise
  be a deletion-only diff and pylint scores messages per statement: the move alone measures
  9.70 -> 9.66 and fails `ci / Lint`.
- API surface: `arch_from_id` and `get_host_arch` are no longer attributes of the
  `archinfo.arch` submodule. An AST walk over every parseable `.py` and `.pyi` file in the
  angr, cle, pyvex, claripy, pypcode and angr-management checkouts finds no
  `from archinfo.arch import` of either name and no `archinfo.arch.` attribute use of
  either; every caller uses `archinfo.arch_from_id`.
- Hooks: `pre-commit run --files` over the four changed files — passed, tree unchanged.
- Docs: `python -m sphinx -b html docs` at master and at this head emits the same 12 warnings,
  character for character apart from the source path; `arch_from_id` is still on `api.html`.
- Downstream, with this branch first on `PYTHONPATH`: `import angr` succeeds and `CFGFast`
  on `binaries/tests/x86_64/fauxware` finds 40 functions.

Not run at this head, for the same reason as before: the complete workspace gate, which
enters `nix develop` and would rewrite native libraries a running corpus sweep has mapped.
CI has not seen this head yet either.

zardus:

This draft and angr/cle#740 overlap in a way that depends on merge order, so flagging it before either lands.

cle#740 makes `ELF._get_compatible_pcode_languages` apply the header fields a Ghidra opinion actually constrains. Two things were wrong there: `secondary` was compared against `e_type` when the opinions document it as `e_flags`, and `int()` cannot parse the hex and bitmask forms, so those opinions fell into a `ValueError` branch that kept them unconditionally. Separately, 47 of the 80 ELF opinions carry no `endian`, and nothing else held the selection to the file's `EI_DATA`. Scored against the ELF identification headers of the objects in a large corpus, that change corrects 11,069 objects and makes none worse: 5,740 PA-RISC objects go from failing to load at all to loading, and 5,329 LoongArch objects move from `lp64f` to `lp64d`, which their `e_flags` of `0x43` says is correct.

The interaction is with the fallback this PR adds. `ELF.extract_arch` consults the p-code opinions only after `arch_from_id` raises. With `ArchPcode.from_id` reachable from `arch_from_id`, machines whose names normalize cleanly — PA-RISC among them — resolve before the opinion path is ever tried, so that path becomes unreachable for several of the machines it exists to serve. `from_id` ranks on processor name, width and endness, so the `e_flags` disambiguation would be silently lost: LoongArch would go back to whichever of `lp64d` and `lp64f` sorts first, regardless of what the file says.

Merged the other way round the two are complementary, and this PR still does useful work for machines the opinions do not cover.

So it seems like one of two things: either this fallback should not preempt the loader for machines an ELF opinion already covers, or `ELF.extract_arch` should consult the opinions before `arch_from_id` rather than after. I do not have a strong view on which is better placed, and it is your PR — but whichever it is, it is easier to decide now than to notice later, because nothing fails loudly when the disambiguation stops running.

The RISC-V nondeterminism this PR fixes incidentally is real and worth keeping: `pypcode.Arch.enumerate()` yields in directory order, and creating that language directory repeatedly with identical contents produced both orders roughly evenly, so a reinstall can change how an RV32 binary decodes. `from_id`'s ranking makes that deterministic. Note that simply sorting the enumeration would not — `andestar_v5.ldefs` sorts first, so plain RV32 objects would deterministically get the vendor superset. The remaining ambiguity that no header field can settle is filed as angr/cle#741.

twizmwazin:
What is the motivating case for this? If we do care to support this, why not make it simple and allow full language ids to be used, and drop all of the complex search and normalization code. Either it matches the name directly or it doesn't exist.

zardus:

You are right on the second half, and I have made that change: the search and normalization are gone, and `arch_from_id` now takes a full language id or nothing. Head `9c6ba269e90646649ec4bf4e2e4a326e6ac12d51`, and the description and validation comment are rewritten to match.

The motivating case is not the one the old description gave, which is the part I got wrong. I went and looked at what the callers actually pass: cle's ELF backend passes pyelftools' `EM_*` name, the PE backend passes pefile's `IMAGE_FILE_MACHINE_*`, COFF, TE and Mach-O pass short literals like `AMD64` and `X64`, and `get_host_arch` passes a uname string. Not one of those 236 identifiers is a sleigh language id, so exact matching gives the loaders nothing at all — and the normalization that did give them something was doing the loader's job in a place that cannot see the file.

What does pass a full language id is anything that stores an architecture by name and reads it back, because `ArchPcode.name` *is* the language id. angr's database serializer is the concrete one: `angr/angrdb/serializers/loader.py` writes `{"name": arch.name, "endness": ..., "bits": ...}` and reloads it with `arch_from_id`. Through that encoder and decoder unmodified:

| Stored `arch.name` | Before | After |
| --- | --- | --- |
| `pa-risc:BE:32:default` | `ArchNotFound` | `ArchPcode` `pa-risc:BE:32:default` |
| `ARM:LE:32:v7` | `ArchARMEL` | `ArchPcode` `ARM:LE:32:v7` |
| `MIPS:BE:32:default` | `ArchMIPS32` | `ArchPcode` `MIPS:BE:32:default` |

70 of the 187 ids land in that second row — `arch_id_map`'s regexes match the processor name inside the id, and the stored endness and bits then let it resolve — so saving a p-code project and reloading it quietly swapped the engine. That is the one place the change is not purely additive: an id answers before `arch_id_map` rather than after it. `arch=` on `Project` and `cle.Loader` reaches the same call, so naming a language there works now too.

On dropping the search rather than defending it, I measured it against `ELF._get_compatible_pcode_languages`, which cle already consults when `arch_from_id` raises, over every `ENUM_E_MACHINE` value at both ELF classes and both byte orders — 752 combinations. The search resolved 3 that the opinions did not: `EM_PARISC` big-endian 32, `EM_SPARC` big-endian 64, and `EM_BPF` at 32-bit class, which no real eBPF object uses. It preempted 13 that the opinions did cover and answered 6 of those differently, LoongArch among them, where the opinion turns on `e_flags` that archinfo cannot see. And the opinions resolve 23 it never reached, including `EM_68K`, `EM_SH`, `EM_MSP430` and `EM_SPARCV9`. It also cost 12.7 ms per miss on the path the ELF backend takes for every unregistered machine, against 0.009 ms now.

PA-RISC is the case I originally built this around, and it belongs in cle. `tests/hppa/test-instr_hppa` has `e_flags` `0x210`, which is exactly the `secondary` constraint on Ghidra's PA-RISC ELF opinion — cle compares that field against `e_type` today, so the opinion never applies. angr/cle#740 fixes it, and with that fix the loader selects `pa-risc:BE:32:default` from the header itself. I have marked the end-to-end PA-RISC claim and the corpus rows in the validation comment superseded rather than leaving them standing.

Still a draft on #363: accepting any language id means every one of the 187 has to construct, and the 8051's one-byte stack pointer raises without that fix.

`python -m pytest tests` — 24 passed; the three new tests fail on the baseline. Details in https://github.com/angr/archinfo/pull/365#issuecomment-5233536803.

archinfo/arch_pcode.py
```diff
@@ -151,9 +152,30 @@ def pcode_arch(self) -> "ArchPcode":
         """
         return self
 
+    @classmethod
+    def from_id(cls, ident: str) -> "ArchPcode | None":
```
twizmwazin: This function seems to be a duplicate from the constructor? Can it be removed and the constructor used instead?

archinfo/arch.py
```diff
@@ -863,12 +863,28 @@ class ArchNotFound(Exception):
...
     """
     Take our best guess at the arch referred to by the given identifier, and return an instance of its class.
 
     You may optionally provide the ``endness`` and ``bits`` parameters (strings) to help this function out.
+
+    A full sleigh language id, such as ``pa-risc:BE:32:default``, names one of the languages pypcode
+    ships and returns an ArchPcode for it. The id already carries its own endness and width, so the
+    ``endness`` and ``bits`` hints do not apply to it.
     """
+    if ":" in ident:
```
twizmwazin: Let's fold the pcode guard in here by doing `and pcode is not None`, and remove the giant comment.

twizmwazin (changes requested):

archinfo/arch_pcode.py
zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

You are right, and it duplicated more than the constructor: `_get_language_by_id`, which the constructor already calls when it is handed a `str`, is the same enumerate-and-match loop. Removed. `arch_from_id` now does

```python
with contextlib.suppress(ArchError):
    return ArchPcode(ident)
```

which covers both cases the classmethod's `None` return covered, because `ArchPcode.__init__` raises `ArchError` for either: no language with that id, and archinfo installed without the `pcode` extra. `archinfo/arch_pcode.py` is back to master's content, so this PR no longer touches that file at all.

One behavioural consequence, since `from_id` lowercased both sides and `_get_language_by_id` compares exactly: `PA-RISC:be:32:DEFAULT` no longer resolves. That is what you asked for last time — either it matches or it does not exist — so I dropped the assertion that covered it rather than making the constructor case-insensitive. Nothing else moved: over the 492 identifiers x 7 endness/width hints in the validation comment, all 3444 resolutions are identical to the head you reviewed.

archinfo/arch.py
zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Done. Removing `from_id` removed the nested `is not None` guard you wanted folded in, so there is one `if` left, and the giant comment is down to the one thing that is not visible in the code:

```python
    if ":" in ident:
        # A language id names one language, so it answers before arch_id_map, whose regexes would otherwise
        # claim ARM:LE:32:v7 for ArchARMEL. The import is delayed because arch_pcode imports this module.
        from .arch_pcode import ArchPcode  # pylint: disable=import-outside-toplevel

        with contextlib.suppress(ArchError):
            return ArchPcode(ident)
```

I kept those two lines because the ordering is load-bearing and silent when it is wrong: move the branch below `arch_id_map` and `ARM:LE:32:v7` comes back as `ArchARMEL` with no error. `test_arch_from_id_by_language_id_of_a_registered_arch` is what fails if it moves. The docstring paragraph is two lines shorter as well.

zardus:

First, the push at 01:39 today came after your review and did not answer it, which I should not have done. It was only a typecheck fix and unrelated to either of your points: rebasing this branch onto #367 left `arch_pcode.py` carrying both master's `if TYPE_CHECKING: import pypcode` and this branch's `pypcode = None` fallback, `ci / Typecheck` went red on the rebased head, and that commit fixed exactly that.

Both of your points are taken, in `20d78c28173006eda13b6b60cb4fc901b5fdfda5`. `ArchPcode.from_id` is gone and `arch_from_id` calls the constructor; the guard is folded and the comment is cut. That also removed the typecheck fix above, along with everything else this branch did to `arch_pcode.py` — the file is back to master's content, and the whole change is now 13 added lines in `arch.py` plus tests.

On the draft: it stays parked until #363 merges. This branch carries #363's commit because accepting any language id means all 187 have to construct, and the 8051's one-byte stack pointer raises without that fix. archinfo squash-merges, so merging this one first would fold #363's commit into it and leave #363 open with a diff that no longer applies. Nothing references this PR, so no other branch is silently being tested against archinfo master in the meantime. Once #363 lands I rebase, which drops the duplicate, and mark this ready.

Validation refreshed in place: https://github.com/angr/archinfo/pull/365#issuecomment-5233536803

zardus:

Out of draft. #363 merged on 2026-08-27 as `f92307b32e138de45b6ed4e8ca7794b67fdff6f8`, and it was the only thing this branch was parked on.

I ran the retarget sequence exactly as the validation comment recorded it. `git rebase --onto origin/master 73a5917 cfa6d72` dropped the duplicated #363 commit, and `git range-diff bf85c7e..cfa6d72 f92307b..d4e902b` reports `73a5917 < -` for the duplicate and `cfa6d72 = d4e902b` for this PR's own commit — unchanged, so the reviewed content is what is here. New head `d4e902b96368accacaad88c51b0dfb63e1cfa734`. The diff against master is now 13 added lines in `archinfo/arch.py` plus tests, and `archinfo/arch_pcode.py` is not touched at all, which the duplicate commit was previously obscuring.

Re-verified on the rebased head before promoting it, and not only re-read:

- `python -m pytest tests` — 26 passed, the same as the focused row in the validation record
- `pre-commit run --files archinfo/arch.py tests/test_pcode.py` — passed, tree unchanged
- Both points from the CHANGES_REQUESTED review are on this head and unmoved by the rebase, since the commit is byte-identical: `ArchPcode.from_id` is gone and `arch_from_id` calls the constructor under `contextlib.suppress(ArchError)`, and the guard is folded into the single `if ":" in ident:` with the comment down to two lines
- The ordering interaction I raised against cle 740 does not apply to this head either. Matching is exact language ids only, and nothing a loader passes is a language id, so `ELF._get_compatible_pcode_languages` is never preempted

Validation refreshed in place: https://github.com/angr/archinfo/pull/365#issuecomment-5233536803

zardus:

`arch_from_id` on sleigh language ids, and the round trip angr's database serializer makes, before and after this change.

**Before** — no language id round-trips; `ARM:LE:32:v7` silently comes back as `ArchARMEL`:

<details><summary>archinfo master</summary>

```
archinfo: archinfo at master 7e72c6c

arch_from_id('pa-risc:BE:32:default')  -> ArchNotFound: Can't find architecture info for architecture pa-risc:be:32:default with 32 bits and unsure endness
arch_from_id('ARM:LE:32:v7')  -> ArchARMEL(name='ARMEL', bits=32, memory_endness=<Endness.LE: 'Iend_LE'>)
arch_from_id('6502:LE:16:default')  -> ArchNotFound: Can't find architecture info for architecture 6502:le:16:default with '' bits and unsure endness
arch_from_id('x86:LE:16:Real Mode')  -> ArchX86(name='X86', bits=32, memory_endness=<Endness.LE: 'Iend_LE'>)

store arch.name and reload it with arch_from_id, over all 187 pypcode languages:
    round-trips to the same ArchPcode : 0
    comes back as a different class   : 70
    raises                            : 117
    first ten that come back as a different class:
        RISCV:LE:64:default                -> ArchRISCV64
        x86:LE:32:default                  -> ArchX86
        x86:LE:32:System Management Mode   -> ArchX86
        AARCH64:LE:64:v8A                  -> ArchAArch64
        AARCH64:BE:64:v8A                  -> ArchAArch64
        AARCH64:LE:64:AppleSilicon         -> ArchAArch64
        MIPS:BE:32:default                 -> ArchMIPS32
        MIPS:LE:32:default                 -> ArchMIPS32
        MIPS:BE:32:R6                      -> ArchMIPS32
        MIPS:LE:32:R6                      -> ArchMIPS32
    first ten that raise:
        RISCV:LE:32:AndeStar_v5            -> ArchNotFound
        RISCV:LE:32:default                -> ArchNotFound
        tricore:LE:32:default              -> ArchNotFound
        tricore:LE:32:tc29x                -> ArchNotFound
        tricore:LE:32:tc172x               -> ArchNotFound
        tricore:LE:32:tc176x               -> ArchNotFound
        M16C/80:LE:16:default              -> ArchNotFound
        M16C/60:LE:16:default              -> ArchNotFound
        Hexagon:LE:32:default              -> ArchNotFound
        NDS32:BE:32:default                -> ArchNotFound

identifiers that are not language ids, unchanged either way:
    arch_from_id('x86') -> ArchX86
    arch_from_id('amd64') -> ArchAMD64
    arch_from_id('arm') -> ArchARMEL
    arch_from_id('mips32') -> ArchMIPS32
    arch_from_id('ppc64') -> ArchPPC64
    arch_from_id('s390x') -> ArchS390X
    arch_from_id('riscv64') -> ArchRISCV64
```

</details>

**After** — all 187 round-trip, and identifiers that are not language ids resolve exactly as before:

```
archinfo: archinfo at this branch, head 26998a9

arch_from_id('pa-risc:BE:32:default')  -> ArchPcode(name='pa-risc:BE:32:default', bits=32, memory_endness=<Endness.BE: 'Iend_BE'>)
arch_from_id('ARM:LE:32:v7')  -> ArchPcode(name='ARM:LE:32:v7', bits=32, memory_endness=<Endness.LE: 'Iend_LE'>)
arch_from_id('6502:LE:16:default')  -> ArchPcode(name='6502:LE:16:default', bits=16, memory_endness=<Endness.LE: 'Iend_LE'>)
arch_from_id('x86:LE:16:Real Mode')  -> ArchPcode(name='x86:LE:16:Real Mode', bits=16, memory_endness=<Endness.LE: 'Iend_LE'>)

store arch.name and reload it with arch_from_id, over all 187 pypcode languages:
    round-trips to the same ArchPcode : 187
    comes back as a different class   : 0
    raises                            : 0

identifiers that are not language ids, unchanged either way:
    arch_from_id('x86') -> ArchX86
    arch_from_id('amd64') -> ArchAMD64
    arch_from_id('arm') -> ArchARMEL
    arch_from_id('mips32') -> ArchMIPS32
    arch_from_id('ppc64') -> ArchPPC64
    arch_from_id('s390x') -> ArchS390X
    arch_from_id('riscv64') -> ArchRISCV64
```

</details>

archinfo/arch.py
```diff
@@ -863,12 +864,23 @@ class ArchNotFound(Exception):
...
 
     You may optionally provide the ``endness`` and ``bits`` parameters (strings) to help this function out.
+
+    A full sleigh language id, such as ``pa-risc:BE:32:default``, returns the ArchPcode for that language. It
+    carries its own endness and width, so the ``endness`` and ``bits`` hints do not apply to it.
     """
+    if ":" in ident:
+        # A language id names one language, so it answers before arch_id_map, whose regexes would otherwise
+        # claim ARM:LE:32:v7 for ArchARMEL. The import is delayed because arch_pcode imports this module.
+        from .arch_pcode import ArchPcode  # pylint: disable=import-outside-toplevel
```
twizmwazin: Let's make this import optional like the others in this file

zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Done in `3c2f1f6`. `pypcode` is now declared at the top of `arch.py` in the same shape as `_pyvex`, `_unicorn`, `_capstone` and `_keystone`, and the branch reads the sentinel:

```python
try:
    import pypcode as _pypcode
except ImportError:
    _pypcode = None
```

```python
    if _pypcode is not None and ":" in ident:
        # A language id names one language, so it answers before arch_id_map, whose regexes would otherwise
        # claim ARM:LE:32:v7 for ArchARMEL. arch_pcode imports this module, so it cannot be imported at the top.
        from .arch_pcode import ArchPcode  # pylint: disable=import-outside-toplevel

        with contextlib.suppress(ArchError):
            return ArchPcode(ident)
```

Without the `pcode` extra installed, `arch_from_id` never enters the branch and answers exactly as it did before language ids were accepted at all: `arch_from_id("x86")` is `ArchX86`, `arch_from_id("pa-risc:BE:32:default")` raises `ArchNotFound`, and `archinfo` still imports with no runtime dependencies. The `contextlib.suppress` now covers only "no language with that id" instead of doubling as the missing-dependency path.

The `from .arch_pcode import ArchPcode` line itself stays inside the function, and I would rather say why than leave it looking unaddressed. `arch_pcode` imports this module, so hoisting it into a top-level `try`/`except ImportError` catches the circular import rather than a missing dependency:

```
ImportError: cannot import name 'Arch' from partially initialized module 'archinfo.arch'
(most likely due to a circular import)
```

That binds the name to `None` and silently disables the feature on a machine where pypcode is in fact installed. `Arch.pcode_arch` on master defers the same import for the same reason.

Regression: `test_arch_from_id_without_pypcode` moved out of the `skipUnless(pypcode is not None)` class so it runs in both configurations, and it now exercises each guard on its own. Running the suite with a `find_spec` hook that makes `pypcode` genuinely unimportable, both heads on the same base:

- `3c2f1f6` — 16 passed, 14 skipped, and `test_arch_from_id_without_pypcode` PASSED
- the head you reviewed — 15 passed, 15 skipped, and that same test SKIPPED, so the without-pypcode claim had never run in the configuration it describes

With pypcode installed both heads give 30 passed, 34 subtests passed, and the before/after capture in the output comment is byte-identical, so nothing a user with the extra sees has moved. Lint and type per changed file against the merge base `b8ffe85`: pylint `archinfo/arch.py` 9.70 -> 9.70 and `tests/test_pcode.py` 10.00 -> 10.00; pyright badness 0.2218 -> 0.2178 and 0.0 -> 0.0.

The branch was also rebased onto `b8ffe85` on the way past; `git range-diff` reports the original commit unchanged.

twizmwazin: Remove the local import since there is a top-level one

zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

The top-level import you are looking at is under `if TYPE_CHECKING:`, so it has no runtime binding, which is why this one is not redundant. I measured all three ways of acting on the comment rather than assuming:

- **Delete this line and keep the `TYPE_CHECKING` one.** `import archinfo` still succeeds, and then `arch_from_id("ARM:LE:32:v7")` raises `NameError: name 'ArchPcode' is not defined`.
- **Promote the `TYPE_CHECKING` import to a real module-level one.** The package stops importing at all: `ImportError: cannot import name 'Arch' from partially initialized module 'archinfo.arch' (most likely due to a circular import)`. `arch_pcode` does `from .arch import Arch, Endness, Register`, and at that point `Arch` is not defined yet.
- **A real import at the bottom of `arch.py`, below the classes.** This one does work — `archinfo.arch.ArchPcode` is bound at runtime, everything resolves, and pylint stays at 9.70. It costs `# noqa: E402` and `# pylint: disable=wrong-import-position`, and it makes `Arch.pcode_arch`'s own deferred import redundant, which is master's line rather than this PR's.

So I have kept the import. It is what `Arch.pcode_arch` already does in this same file, under the comment `# Delayed import to avoid circular dependency`. For completeness, `arch.py` names `ArchPcode` in three places — the `TYPE_CHECKING` import, `pcode_arch`'s runtime import, and this one — and only this one belongs to this PR.

What `dd03d23` does change is say that in the code, so the next reader does not have to ask. The `TYPE_CHECKING` import now carries `# Typing only. arch_pcode imports this module, so its runtime import is delayed to each use site.`, and this site reuses `pcode_arch`'s wording: `Delayed import to avoid circular dependency, as in Arch.pcode_arch.`

Suites at `dd03d23`, unchanged from the previous head: 30 passed and 34 subtests with pypcode installed; 16 passed, 14 skipped with `pypcode` made genuinely unimportable by a `sys.meta_path` finder whose `find_spec` raises, the block asserted to fire before pytest starts. Lint and type per changed file against `b8ffe85`: `archinfo/arch.py` 9.70 -> 9.70, `tests/test_pcode.py` 10.00 -> 10.00; pyright badness 0.22175 -> 0.21762 and 0.0 -> 0.0.

If you would rather have the bottom-of-file import despite the two suppressions, say so and I will push it.

zardus:

This pull request is green on its own, but in the [angr/mono#3](https://github.com/angr/mono/pull/3) rollup — 160 open pull requests merged onto upstream head and run through one CI — it is the reason six job families are red. Reporting it here because the fix is a rebase, not a change to any of the pull requests that fail.

**What happens.** #365 and #375 both rewrite the signature and the opening docstring paragraph of `arch_from_id` in `archinfo/arch.py`:

```
<<<<<<< #365
def arch_from_id(ident: str, endness: str = Endness.ANY, bits: int | str = "") -> Arch:
=======
def arch_from_id(ident: str, endness=Endness.ANY, bits: str | int = "") -> Arch:
>>>>>>> #375
```

They are merged in number order, so #365 lands and #375 is skipped with `conflicts with the applied set: archinfo/arch.py`. #375 is the base of a four-repository stack — angr/binaries#206, angr/pyvex#576, angr/cle#795, angr/angr#6982 — and the other four apply anyway, because nothing drops a pull request whose dependency was skipped at assembly time rather than excluded at selection time. The result is a tree where `angr/calling_conventions.py` and `cle/backends/elf/elf.py` reference an `archinfo.ArchMIPSN32` that the tree does not contain.

**What it costs.** `import angr` fails outright:

```
File ".../angr/calling_conventions.py", line 2602, in SimCCN32
    ARCH = archinfo.ArchMIPSN32
AttributeError: module 'archinfo' has no attribute 'ArchMIPSN32'. Did you mean: 'ArchMIPS32'?
```

That single line reddens all five `pyinstaller` jobs (ubuntu-22.04 / ubuntu-24.04 / ubuntu-24.04-arm / windows-2022 / macos-15), the whole `angr-management` suite on every platform, and — through the loader — 5 of 7 tests in cle's `tests/test_mips_n32.py` in the `pyodide · pyvex` job. The identical base with no pull requests applied ([run 33233551760](https://github.com/angr/mono/actions/runs/33233551760)) is green on every one of them, so this is an interaction and not mono infrastructure.

**Why this side.** #365 carries no failing test of its own; it is the pull request whose presence blocks the other. Re-running the merges at heads `dd03d233` (#365) and `39bab86b` (#375) on base `b8ffe857`: #375 merges cleanly with #364 and with #376 and conflicts only with #365, and the set `364 + 375 + 376` merges clean, passes archinfo's own 35 tests, and turns cle's `test_mips_n32.py` from 5 failures into 7 passes. So #365 is excluded from the rollup selection until one side rebases; the conflict is textual, not behavioural, and either rebase retires the entry.

session: sharpen

zardus:

How much of a corpus's failure surface this removes, measured rather than
argued.

**Sample.** 12,000 objects drawn uniformly at random, from a seeded permutation,
out of a 624,920-object internal corpus of compiler- and vendor-produced
binaries; 11,989 were retrievable and probed. Rates below carry 95% Wilson
intervals.

**Method.** Each object is loaded with the catalogue's declared load recipe and
`CFGFast` is taken as far as the failure under test; the same object set is then
probed against `master` and against this branch's head `dd03d233` in one
environment, so before and after are the same objects rather than two samples.
A second pass runs `CFGFast` to completion on a random subsample of the objects
this branch clears.

**Before.** 1,212 / 11,989 = **10.11%** (CI 9.58–10.66) fail with `ArchNotFound`
raised from `arch_from_id`. It is the single largest failure class in the
sample; the next is 6.57%.

**After.** 951 of those 1,212 no longer fail there — **7.93%** of the sample
(CI 7.46–8.43). Running `CFGFast` to completion on a random 250 of them, 207
finish (82.8%, CI 77.6–87.0); the other 43 fail later, in the lifter and in
pyvex, not in architecture resolution. 261 still fail at `arch_from_id`: 211 for
identifiers that are not sleigh language ids at all (`sharp-sm83`, `ricoh-5a22`,
`em_sparc`, `em_parisc`), and 50 that now get past resolution and fail elsewhere
in cle.

The cleared population is embedded and retro firmware: resolved architectures
include z80 (93), 6502 (79), RISCV (59), PIC-18 (42), HC08 (38), 80390 (37),
HCS08 (37), z180 (24), 8051 (19), 65C02 (17), 68000 (9). One 32-bit Xtensa image
that raises `ArchNotFound` on master recovers 8,086 functions on this head.

A 500-object control set that already reached CFG on master is unchanged on this
branch — 0 of 500 differ.

The corpus is not redistributable, so the objects are described by architecture,
format and OS rather than named; none of the affected objects is byte-identical
to anything tracked in `angr/binaries`.

**Still current at head `26998a9`, 2026-08-30.** The branch was rebased onto archinfo
master `7e72c6c` after #375 merged; the only line of the branch's own diff that changed
there is the `bits: str | int` annotation on `arch_from_id`, which master now carries. On
review the resolution then moved out of `arch.py` into `__init__.py`. Neither change moved
an answer: over 548 identifiers x 7 endness/width hint combinations, `arch_from_id` on
`26998a9` is byte-identical to `14c81fc`, and with pypcode installed — which is how these
rates were probed — the old `_pypcode is not None and ":" in ident` and the new
`":" in ident` are the same condition. So the numbers above still describe this head.

**Still current at head `c24c173`, 2026-09-05.** The branch was rebased onto archinfo
master `384c676` and the resolution moved from a wrapper in `__init__.py` to the whole
`arch_from_id` there, with `get_host_arch` alongside it. Neither moved an answer: over 438
identifiers x 7 endness/width hint combinations — 3,066 resolutions, all 187 language ids
among them — this head is byte-identical to `26998a9` replayed on the same base. So the
rates above still describe it. The one code change beyond the move is `register_arch`'s
`re._pattern_type` repair, which no corpus object reaches: archinfo registers with string
patterns, which short-circuit before that check.

session: sharpen

zardus:

Rebased onto archinfo master `7e72c6c`; the branch was CONFLICTING after #375 merged. New head `14c81fc`.

The conflict was `arch_from_id`'s signature and docstring in `archinfo/arch.py`. #375 annotated `bits: str | int` and rewrote the sentence describing `bits`; this branch had annotated both `endness` and `bits` itself and rewritten the same lines. Resolved by taking master's `bits: str | int` and master's docstring wording, keeping this branch's `endness: str` annotation and the paragraph it adds about full sleigh language ids. So the only signature change the branch still carries is the `endness` annotation.

That annotation is load-bearing rather than cosmetic: `Endness` is a `StrEnum`, so with `endness` unannotated pyright infers the parameter as `Endness` from its default, and this branch's own `test_arch_from_id_by_language_id` passes the literal `"le"`. Dropping it took `tests/test_pcode.py` from badness 0.0 to 0.095 in the merge-base type comparison. With it kept, both changed files are clean.

Otherwise the change did not move: comparing the branch's own diff before and after the rebase with context discarded, the signature line is the only difference.

Validation at the new head: `pytest tests` — 37 passed, 34 subtests. `arch_from_id("pa-risc:BE:32:default")` returns the ArchPcode and `arch_from_id("Xtensa:BE:32:default", "le", 64)` still answers `Xtensa:BE:32:default`, so the id keeps winning over the hints. `import angr` and `archinfo.ArchMIPSN32` both work with this branch applied on top of #375, which is what the two used to disagree about. The merge-base lint and type comparison reports no regression on either changed file.

twizmwazin:
Can you reduce the import nonsense by putting this in init.py?

zardus:

Done in `26998a9`: the resolution now lives in `__init__.py`, where `ArchPcode` is already imported at module level, so `arch.py`'s whole diff against master is the `endness: str` annotation. No answer changed — the 3,836-resolution dump in the validation record is byte-identical to the one at `14c81fc`.

twizmwazin:
Put the whole is_arch_arm function there, and not below the __all__ decl

zardus:

Done in `c24c173`, on a branch rebased onto master `384c676`.

`arch_from_id` is now the whole function in `__init__.py`, above the `__all__` list — no wrapper, no delegation back into `arch.py`. `get_host_arch` moved with it, because it is one call to `arch_from_id` and leaving it behind would have needed a deferred import back into the package root, which is the shape you asked me to get rid of.

That closes the two inline comments as well. This change no longer adds an import to `arch.py` at all: `__init__.py` already imports `ArchPcode` at module level, so the branch just calls it, and there is no optional-import guard either. The one deferred `from .arch_pcode import ArchPcode` still in the file is master's, in `Arch.pcode_arch` at `arch.py:497`, and this branch does not touch it.

One thing came with the move that I would rather flag than bury. Without a compensating repair `arch.py` would be a deletion-only diff, and pylint scores a file by messages per statement, so `ci / Lint` fails it at 9.70 -> 9.66 with nothing added to it. The repair is the file's only pylint error, and a real defect: `register_arch`'s type check reads `re._pattern_type`, which Python removed in 3.7. A list of plain strings short-circuits before reaching it, which is why archinfo's own 21 registrations work; any list holding something else raised `AttributeError`, both the compiled patterns the docstring documents and the bad types the check exists to reject. `re.Pattern` is the name that exists, the annotation and the docstring now say what the check does, and `tests/test_register_arch.py` covers both directions and fails on master. `arch.py` goes 9.70 -> 9.77. Say the word and I will split it into its own pull request.

`arch_from_id` and `get_host_arch` are no longer attributes of the `archinfo.arch` submodule. Nothing in angr, cle, pyvex, claripy, pypcode or angr-management imports either from there; every caller uses `archinfo.arch_from_id`.

Nothing else moved. Over a pool of 438 identifiers — every `all_arches` name, every pyelftools `ENUM_E_MACHINE` name, every pefile `IMAGE_FILE_MACHINE_*` name, all 187 sleigh language ids, and 22 literal identifiers of which 14 are not already in those sets — at 7 endness/width hint combinations, 3,066 resolutions, the answers are byte-identical to the head you last reviewed replayed on this base. The same dump against master differs in 1,309 rows, every one of them a language id, and none goes from resolved to raising. `pytest tests`: 46 passed, 34 subtests. With `pypcode` made genuinely unimportable by a `sys.meta_path` finder, `arch_from_id("x86")` is `ArchX86`, `get_host_arch()` answers, and the PA-RISC id raises `ArchNotFound` — unchanged. Merge-base lint and type, per changed file: `archinfo/__init__.py` 10.00 -> 10.00 and pyright errors 0 -> 0; `archinfo/arch.py` 9.70 -> 9.77 and 21 -> 20; `tests/test_pcode.py` 10.00 -> 10.00 and 0 -> 0; `tests/test_register_arch.py` new, 10.00 and 0.

Validation record refreshed at this head: https://github.com/angr/archinfo/pull/365#issuecomment-5233536803

archinfo/arch.py
zardus: THIS MESSAGE WAS GENERATED BY AN AUTOMATED PROCESS

Both of these are resolved by the move rather than by a change here: `arch_from_id` is now the whole function in `__init__.py`, so this branch adds no import to `arch.py` at all. There is no optional-import guard to make optional and no local `ArchPcode` import to remove, because `__init__.py` already imports `ArchPcode` at module level.

The one deferred `from .arch_pcode import ArchPcode` still in the file is master's, inside `Arch.pcode_arch` at `arch.py:497`, and this branch does not touch it.

Detail in https://github.com/angr/archinfo/pull/365#issuecomment-5555324565

PR #373, opened 2026-08-25
Let ARM express big-endian data with little-endian instructions

zardus:

ARM BE8 is big-endian data with little-endian instructions, which is the normal layout for modern big-endian ARM. No `Arch` can express it: `ArchARM.__init__` sets `instruction_endness` to little-endian only when the memory endness is little-endian, so `ArchARM(Endness.BE)` inherits the base class default and claims big-endian instructions. Its byte-reversed prologue patterns follow from the same assumption, so the prologue scan looks for the wrong bytes as well.

`instruction_endness` becomes a constructor argument, defaulting to the memory endness so nothing changes for any existing caller. When it differs, the architecture keeps the little-endian prologue and epilogue sets and the little-endian capstone, keystone and unicorn modes, which is what BE8 actually is.

I chose this over a separate `ArchARMBE8` class deliberately. BE8 is orthogonal to the float ABI — every object I have seen is `ArchARMHF` — so a distinct class means one per combination, and a new `name` would fall out of every name-keyed table downstream in angr: the calling-convention maps, `cfg_arch_options`, the indirect-jump resolvers, the ccall rewriters, the disassembly helpers. `ArchARMHF(BE, instruction_endness=LE)` stays an `ARMHF` to all of them. Say the word if you would rather have the class and I will do it that way.

One consequence worth flagging: `Arch.__eq__` and `__hash__` were keyed on name, bits and memory endness, so a BE8 architecture compared equal to a BE32 one and would have collided in any architecture-keyed cache. They now include `instruction_endness`. Over all 22 registered architectures the resulting hash partition is byte-identical to before and there are no new equality collisions, because for every architecture that exists today `instruction_endness` is a function of the class and the memory endness.

There is deliberately no new `arch_from_id` spelling. There is no established name for this configuration, and the only place the fact exists is the ELF header, which is where the open ARM-BE8 request in cle reads it.

Scope, stated plainly: this makes the representation correct, and capstone and the p-code lifter then decode BE8 correctly. **VEX still cannot lift it.** libVEX drives instruction fetch and data from a single endness value, so `instruction_endness` never reaches it — see angr/pyvex#565. This is a prerequisite for that work rather than a fix for it.

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -324,12 +324,17 @@ class Arch:
         return f"<Arch {self.name} ({self.memory_endness[-2:]})>"
 
     def __hash__(self):
-        return hash((self.name, self.bits, self.memory_endness))
+        return hash((self.name, self.bits, self.memory_endness, self.instruction_endness))
 
     def __eq__(self, other):
         if not isinstance(other, Arch):
             return False
-        return self.name == other.name and self.bits == other.bits and self.memory_endness == other.memory_endness
+        return (
+            self.name == other.name
+            and self.bits == other.bits
+            and self.memory_endness == other.memory_endness
+            and self.instruction_endness == other.instruction_endness
+        )
 
     def __ne__(self, other):
         return not self == other
--- a/archinfo/arch_arm.py
+++ b/archinfo/arch_arm.py
@@ -50,16 +50,18 @@ class ArchARM(Arch):
     ARM architecture specific subclass
     """
 
-    def __init__(self, endness=Endness.LE):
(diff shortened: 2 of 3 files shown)


zardus:

Closing this as superseded by angr/archinfo#374, which covers the same change against the same files and goes further: it also gives `ArchAArch64` and `ArchRISCV64` a class-level `instruction_endness`, so their big-endian variants decode instead of reporting `Ijk_NoDecode`.

Nothing here is lost — please review angr/archinfo#374 instead.

session: pr-review

PR #374, opened 2026-08-26
Key the VEX guest and encodings on instruction endness, not data endness

zardus:

`Arch` carries an `instruction_endness` ivar that nothing reads: `vex_archinfo["endness"]`, the capstone and keystone modes, the ret and nop encodings and the ARM prologue patterns all derive from the memory endness. Where the two differ — ARM BE8, big-endian data with little-endian instructions — that is why nothing decodes.

`ArchARM` now takes `instruction_endness`, and inside the existing big-endian branch the guest endness, the assembler and disassembler modes, the instruction encodings and the prologue and epilogue patterns key on it rather than on the data endness; it also joins `__hash__` and `__eq__`. The keying is confined to that branch on purpose, since applying it unconditionally would flip little-endian-built architectures to a big-endian guest. `ArchAArch64` and `ArchRISCV64` gain `instruction_endness = Endness.LE` at class level, so their big-endian variants decode rather than reporting `Ijk_NoDecode` for either byte order.

`VexArchInfo` still carries a single endness, so a BE8 guest decodes correctly but tags its data accesses little-endian; separating the two needs a libVEX change. Fixtures: angr/binaries#202. Validation: https://github.com/angr/archinfo/pull/374#issuecomment-5440322048

session: mega-corpus

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -190,18 +190,27 @@ class Arch:
             self.vex_archinfo = _pyvex.enums.default_vex_archinfo()
 
         if endness == Endness.BE:
+            # An architecture may store instruction words in the opposite order to its data, so the
+            # decoders and the instruction byte patterns follow instruction_endness rather than the
+            # memory endness. ARM BE8 is the case that differs; everywhere else the two agree.
+            # VEX carries one endness per guest and uses it both to fetch instructions and to
+            # tag the data accesses it emits, so a BE8 guest decodes correctly while its lifted
+            # loads and stores claim to be little-endian. Expressing both needs a VEX change.
             if self.vex_archinfo:
-                self.vex_archinfo["endness"] = _pyvex.enums.vex_endness_from_string("VexEndnessBE")
+                self.vex_archinfo["endness"] = _pyvex.enums.vex_endness_from_string(
+                    "VexEndnessBE" if self.instruction_endness == Endness.BE else "VexEndnessLE"
+                )
             self.memory_endness = Endness.BE
             self.register_endness = Endness.BE
-            if _capstone and self.cs_mode is not None:
-                self.cs_mode -= _capstone.CS_MODE_LITTLE_ENDIAN
-                self.cs_mode += _capstone.CS_MODE_BIG_ENDIAN
-            if _keystone and self.ks_mode is not None:
-                self.ks_mode -= _keystone.KS_MODE_LITTLE_ENDIAN
-                self.ks_mode += _keystone.KS_MODE_BIG_ENDIAN
-            self.ret_instruction = reverse_ends(self.ret_instruction)
-            self.nop_instruction = reverse_ends(self.nop_instruction)
+            if self.instruction_endness == Endness.BE:
+                if _capstone and self.cs_mode is not None:
+                    self.cs_mode -= _capstone.CS_MODE_LITTLE_ENDIAN
+                    self.cs_mode += _capstone.CS_MODE_BIG_ENDIAN
+                if _keystone and self.ks_mode is not None:
+                    self.ks_mode -= _keystone.KS_MODE_LITTLE_ENDIAN
+                    self.ks_mode += _keystone.KS_MODE_BIG_ENDIAN
+                self.ret_instruction = reverse_ends(self.ret_instruction)
+                self.nop_instruction = reverse_ends(self.nop_instruction)
 
         if self.register_list and _pyvex is not None:
             (_, _), max_offset = max(_pyvex.vex_ffi.guest_offsets.items(), key=lambda x: x[1])
@@ -324,12 +333,17 @@ class Arch:
         return f"<Arch {self.name} ({self.memory_endness[-2:]})>"
 
     def __hash__(self):
-        return hash((self.name, self.bits, self.memory_endness))
+        return hash((self.name, self.bits, self.memory_endness, self.instruction_endness))
 
     def __eq__(self, other):
         if not isinstance(other, Arch):
             return False
-        return self.name == other.name and self.bits == other.bits and self.memory_endness == other.memory_endness
+        return (
+            self.name == other.name
+            and self.bits == other.bits
+            and self.memory_endness == other.memory_endness
+            and self.instruction_endness == other.instruction_endness
+        )
 
     def __ne__(self, other):
         return not self == other
--- a/archinfo/arch_arm.py
+++ b/archinfo/arch_arm.py
@@ -50,16 +50,20 @@ class ArchARM(Arch):
     ARM architecture specific subclass
     """
 
-    def __init__(self, endness=Endness.LE):
-        instruction_endness = None
+    def __init__(self, endness=Endness.LE, instruction_endness=None):
+        if instruction_endness is None:
+            instruction_endness = Endness.LE if endness == Endness.LE else Endness.BE
         if endness == Endness.LE:
-            instruction_endness = Endness.LE
             self.pcode_id = "ARM:LE:32:v7"
+        elif instruction_endness == Endness.LE:
+            # BE8: big-endian data with little-endian instruction words, ARMv6 onward. Ghidra calls
+            # this layout LEBE, and the ELF header marks it with EF_ARM_BE8.
+            self.pcode_id = "ARM:LEBE:32:v7LEInstruction"
         else:
             self.pcode_id = "ARM:BE:32:v7"
 
         super().__init__(endness, instruction_endness=instruction_endness)
-        if endness == Endness.BE:
+        if endness == Endness.BE and instruction_endness == Endness.BE:
             self.function_prologs = {
                 # br"\xe9\x2d[\x40-\x7f\xc0-\xff][\x00-\xff]", # stmfd sp!, {xxxxx, lr}
                 rb"\xe5\x2d\xe0\x04",  # push {lr}
--- a/tests/test_arm.py
+++ b/tests/test_arm.py
@@ -1,0 +1,91 @@
+# pylint:disable=no-self-use
+import unittest
+
+from archinfo import ArchARM, ArchARMEL, ArchARMHF
+from archinfo.arch import Endness
+
+try:
+    import pyvex
+except ImportError:
+    pyvex = None
+
+BX_LR = b"\x1e\xff\x2f\xe1"  # bx lr, as stored in a little-endian instruction stream
+PUSH_LR = rb"\x04\xe0\x2d\xe5"  # push {lr}, likewise
+PUSH_LR_REVERSED = rb"\xe5\x2d\xe0\x04"
+
+
+needs_pyvex = unittest.skipUnless(pyvex is not None, "pyvex not installed")
+
+
+def vex_endness(arch):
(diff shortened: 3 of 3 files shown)


zardus:

Validation record for head `08e5f12953279187ea6a50e68b3b762bbd77469a` against merge base `bf85c7e47bb469878c564480e28c677abbe35acc`.

- Regression: `python -m pytest tests/test_arm.py` — 7 passed on the head. The same file copied onto the merge base gives 4 failed, 3 passed: the four BE8 cases raise `TypeError: ArchARM.__init__() got an unexpected keyword argument 'instruction_endness'`, while the little-endian and BE32 cases, which do not pass the keyword, pass there as well
- Full suite: `python -m pytest tests` — 18 passed, 8 skipped on the head; the skips are `tests/test_pcode.py`, which needs pypcode and is unrelated to this change
- Environment: Python 3.12.14, pytest 9.1.1, pyvex 9.3.3 from PyPI, in a virtualenv created for this worktree so the two lifting assertions ran against a real VEX rather than skipping

Caveats: this is archinfo's own suite only — no workspace-wide gate, and no consumer repository was rebuilt or retested. The behaviour change outside ARM, big-endian AArch64 and RISC-V going from `Ijk_NoDecode` to decoding, has no fixture behind it because `angr/binaries` has no big-endian image for either architecture. The ARM BE8 and BE32 fixtures are in angr/binaries#202, and nothing in this suite loads a file. Remaining limitation, unchanged by this branch: `VexArchInfo` carries one endness, so a BE8 guest's data accesses are still tagged little-endian.

twizmwazin:
Given VEX does not actually support these architectures, can we utilize pypcode+ArchPcode instead for these? Otherwise I don't think these changes are a good idea without first having them sorted out in VEX.

zardus:

Yes — pypcode + `ArchPcode` can express BE8 today, so closing this was right. Unlike the MIPS n32 case, Ghidra models BE8 as a first-class language keyed on `EF_ARM_BE8` and the distinction survives all the way into a running `angr` state. Two small `ArchPcode` bugs and one cle change are still needed, and one half of the VEX premise is wrong; evidence below.

**The language exists and is keyed on the flag.** [`ARM.ldefs:50-65`](https://github.com/angr/pypcode/blob/559aacdc9d363fd19477d9daa40721279cd99248/pypcode/processors/ARM/data/languages/ARM.ldefs#L50-L65):

```xml
<language processor="ARM" endian="big" instructionEndian="little" size="32"
          variant="v8LEInstruction" slafile="ARM8_le.sla" id="ARM:LEBE:32:v8LEInstruction">
  <description>Generic ARM/Thumb v8 little endian instructions and big endian data</description>
```

`ARM:LEBE:32:v7LEInstruction` is the same shape at `ARM.ldefs:132-147`. These are the only two of the 24 ARM ids carrying `instructionEndian` (no LEBE variant for v6/v5/v4/Cortex/v8-m). [`ARM.opinion:44-47`](https://github.com/angr/pypcode/blob/559aacdc9d363fd19477d9daa40721279cd99248/pypcode/processors/ARM/data/languages/ARM.opinion#L44-L47) selects between them on the flag itself:

```xml
<constraint primary="40" processor="ARM" size="32" variant="v8"
    secondary= "0b .... .... 0... .... .... .... .... ...."/>
<constraint primary="40" processor="ARM" size="32" variant="v8LEInstruction"
    secondary= "0b .... .... 1... .... .... .... .... ...."/>  <!-- EF_ARM_BE8 -->
```

`primary="40"` is `EM_ARM`; the secondary matches `e_flags` bitwise, the set bit being 23 = `0x00800000`.

**It survives into pypcode, which is the part I expected to fail.** pypcode never reads `instructionEndian` — zero occurrences in any `.py` or `.cpp` — and `Context.__init__` hands the translator nothing but a path ([`pypcode/__init__.py:242`](https://github.com/angr/pypcode/blob/559aacdc9d363fd19477d9daa40721279cd99248/pypcode/__init__.py#L242)). It does not need to, because the id already names a different `.sla`:

```
139e6973d98ce992cba17a018fe2928cb87d35ccdaf03f8b202545692332fb82  ARM8_le.sla  <- ARM:LE:32:v8 and ARM:LEBE:32:v8LEInstruction
3beeddd890e5a5a16b0ba1bceefe029f222c28c3abb964321e74a6a23e7c38f0  ARM8_be.sla  <- ARM:BE:32:v8
```

Translating `00 00 91 e5` — `ldr r0,[r1]` as BE8 lays it down — gives p-code byte-identical between `LEBE` and `LE`, and different from `BE`:

```
ARM:LEBE:32:v8LEInstruction   ldr r0,[r1,#0x0]     unique[10500:4] = r1 + 0x0 ; r0 = *[ram]unique[10500:4]
ARM:BE:32:v8                  andeq r9,r0,r5, ror #0x3
```

The big-endian *data* half comes from archinfo rather than from SLEIGH: `ArchPcode` reads the ldefs `endian` attribute ([`arch_pcode.py:39`](https://github.com/angr/archinfo/blob/da171ca0dd8eec1e16dd9c4f04576b2c958e2aea/archinfo/arch_pcode.py#L39)), and the p-code engine takes every `LOAD`/`STORE` endianness from `arch.memory_endness`, never from the SLEIGH address space ([`emulate.py:284`](https://github.com/angr/angr/blob/0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1/angr/engines/pcode/emulate.py#L284), [`:303`](https://github.com/angr/angr/blob/0c293dc0dea6a97efd583b7fe9e56912f0e8e2c1/angr/engines/pcode/emulate.py#L303)). End to end that is genuine BE8, with no patch to anything:

```python
a = archinfo.ArchPcode("ARM:LEBE:32:v8LEInstruction")
p = angr.load_shellcode(bytes.fromhex("000091e5"), arch=a, load_address=0x1000)
# disasm: [('ldr', 'r0,[r1,#0x0]')]
s = p.factory.blank_state(addr=0x1000)
s.memory.store(0x8000, bytes.fromhex("11223344")); s.regs.r1 = 0x8000
s.step(num_inst=1).successors[0].regs.r0     # <BV32 0x11223344>
```

Little-endian instruction decode, big-endian data load. This patch was written on the assumption that nothing in the tree could do that, and that assumption was wrong.

**Two bugs it walks into, both in `ArchPcode`, both one-liners.**

1. [`arch_pcode.py:40`](https://github.com/angr/archinfo/blob/da171ca0dd8eec1e16dd9c4f04576b2c958e2aea/archinfo/arch_pcode.py#L40) is `self.instruction_endness = self.endness`, so the one arch that could state the distinction reports `Iend_BE` for a language whose own description says "little endian instructions". `lang.instructionEndian` is already reachable through `ArchLanguage.__getattr__`. It is inert for the p-code lifter but not everywhere: `cfg_fast.py:5794` and `:5816` gate a byte swap on `instruction_endness == Endness.LE`, and `is_arm_arch` matches on `name.startswith("ARM")`, which `"ARM:LEBE:32:..."` does — so CFGFast skips a swap it should perform.
2. `register_endness` comes out `Iend_BE` over a register file built from the *little-endian* sla. `s0=0x300, s1=0x304, d0=0x300`, identical to `ARM:LE:32:v8`, where `ARM:BE:32:v8` has `s0=0x37c, d0=0x3f8`. So every VFP/NEON sub-register alias inverts — storing `0x0011223344556677` to `d0` reads back `s0=0x00112233` instead of `0x44556677`. `r0..r15` don't overlap, so it is invisible until floating point.

Worth noting `AARCH64:BE:64:v8A` does this properly: `instructionEndian="little"` *plus* its own `AARCH64BE.sla` with big-endian overlays (`w0=0x4004`, `s0=0x501c`). ARM's LEBE ids just point at the LE sla, which is why the register half is inconsistent. Happy to send both fixes.

**Nothing dispatches to it, though.** `arch_from_id("EM_ARM", "be", 32)` returns `ArchARM(Iend_BE)`, because [`arch_arm.py:569`](https://github.com/angr/archinfo/blob/da171ca0dd8eec1e16dd9c4f04576b2c958e2aea/archinfo/arch_arm.py#L569) registers `.*arm.*|.*thumb.*` with `Endness.ANY` — `ArchNotFound` is never raised for 32-bit ARM, so cle's p-code fallback at [`elf.py:350-353`](https://github.com/angr/cle/blob/b58ea02a446106647cdaae32bdf91b7062404cc1/cle/backends/elf/elf.py#L350-L353) can never fire for it, and `elf.py:343-346` returns `ArchARMEL`/`ArchARMHF` before that `try` anyway. `arch_pcode.py` never calls `register_arch`, so `arch_from_id` cannot return an `ArchPcode` for anything. Nothing in archinfo or cle reads `EF_ARM_BE8` today. So the explicit `e_flags` branch in angr/cle#793 is needed either way — only its return value changes, from `ArchARM(BE, instruction_endness=LE)` to `ArchPcode("ARM:LEBE:32:v7LEInstruction")`.

**On VEX: half right, and it is the important half.** BE8 is genuinely unrepresentable. `VexArchInfo` carries a single `endness` ([`libvex.h:325`](https://github.com/angr/vex/blob/875f7c9a5f6be621b4f000c29c016e15ddf32207/pub/libvex.h#L325)), and the ARM front end derives *both* instruction fetch and data endianness from it: `getUInt` branches on `guest_endness` ([`guest_arm_toIR.c:204-216`](https://github.com/angr/vex/blob/875f7c9a5f6be621b4f000c29c016e15ddf32207/priv/guest_arm_toIR.c#L204-L216)), and `#define guest_memory_endness (guest_endness == VexEndnessBE ? Iend_BE : Iend_LE)` ([`:130`](https://github.com/angr/vex/blob/875f7c9a5f6be621b4f000c29c016e15ddf32207/priv/guest_arm_toIR.c#L130)). One flag buys the right instructions or the right data, never both — exactly the caveat in this PR's own description, and a good reason not to take the patch.

Big-endian ARM itself *is* supported, though, and has been since [62c35e6](https://github.com/angr/vex/commit/62c35e6a3b013cc85b059c41bdb1d58c72f129c7) ("arm: proper support for big-endian guests", 2017), which made instruction fetch runtime-keyed and deleted upstream's `vassert(vta->archinfo_guest.endness == VexEndnessLE);` from `main_main.c`. There are no hardcoded `Iend_LE` literals left in the ARM decoder; all 21 memory accesses go through the macro. So BE32 lifts today. One live hole for whoever meets it next: `floatGuestRegOffset` still has `vassert(0)` on the big-endian path (`guest_arm_toIR.c:875-892`), so `vmov s0, r0` aborts under a BE guest.

**What happens next.** Not reopening this. angr/cle#793 pins archinfo to this branch, so it cannot merge as written — its options are to return an `ArchPcode` from the `EF_ARM_BE8` branch instead, or to be closed; I'll settle that on that PR rather than here. One trade worth naming before anyone picks: routing BE8 through `ArchPcode` moves it off VEX and capstone onto `UberEnginePcode`, and I have not measured what that costs for this target. If that trade is unattractive, the remaining option is the `VexArchInfo` endness split, which is the "sorted out in VEX" you named. Also: angr/cle#610 is an older third-party attempt at the same problem in the same file, and it predates all of this.

PR #375, opened 2026-08-27
Add ArchMIPSN32 for 64-bit MIPS with 32-bit pointers

zardus:

### Problem

MIPS n32 and O64 put a 64-bit MIPS instruction stream in an ELFCLASS32 container: 64-bit instructions, 32-bit pointers and Elf32_* structures throughout. No architecture in archinfo can say that, because `bits` is both the pointer width and, through `vex_arch`, the instruction set.

On the n32 object `tests/mipsn32/n32_el_dynamic` (`e_flags` `0x80000027`, `EF_MIPS_ABI2` set) the two candidates each get one half right:

```
architecture   bits bytes vex_arch         struct_fmt  0x6c4 decodes to
ArchMIPS32       32     4 VexArchMIPS32    '<I'        <decodes to nothing>
ArchMIPS64       64     8 VexArchMIPS64    '<Q'        sd $gp, 8($sp)
```

`ArchMIPS32` cannot decode the 64-bit `sd` the prologue spills `$gp` with. `ArchMIPS64` strides that object's `.got` — `sh_entsize` 4, at `0x10a30` — by eight, so its first four slots read `0x8000000000000000 0x00010a10000006c0 0x000008b0000109f0 0x00010a2400010000`, each one two real slots glued together.

### Root cause

`Arch.__init__` derives `self.bytes` from `bits`, and `ArchMIPS64` sets `bits = 64` as a class attribute, so the width used to read and write pointers is the same number that selects the VEX guest. There is no separate knob and nothing between the two classes.

### Fix

`archinfo/arch_mips64.py` gains `ArchMIPSN32`, subclassing `ArchMIPS64` and setting `bits = 32`. It inherits the VEX guest, the capstone mode and the 8-byte register file, and narrows the pointer, `sizeof["long"]` and the TLS head to the 32-bit MIPS shape:

```
ArchMIPSN32      32     4 VexArchMIPS64    '<I'        sd $gp, 8($sp)
```

`bits` is the pointer width here and nothing else: the argument registers stay eight bytes, and a consumer that needs a register width reads it from the register file.

It is deliberately not registered for name-based lookup. `ArchMIPS32` already claims every 32-bit MIPS name, and an ABI is not recoverable from a name — detection belongs to the loader, which reads `e_flags`.

### Testing

`tests/test_mips.py` asserts the 32-bit word size and `>I`/`<I` format, the 64-bit VEX guest, that `sp`, `ra`, `pc` and the eight argument registers stay 8 bytes and equal `ArchMIPS64`'s, both `pcode_id`s and the C type sizes. It fails to import on master, which has no `ArchMIPSN32`. The fixture above comes from the binaries half of this work.

These four must land together: with the loader and the architecture definition ahead of the lifter, `MIPSN32` reaches a lifter with no entry for it and recovery on these objects drops to zero blocks. Merge order: archinfo 375, then pyvex 576, then cle 795, then angr 6982.

Validation: https://github.com/angr/archinfo/pull/375#issuecomment-5440321765

session: mega-corpus

### DIFF
--- a/archinfo/__init__.py
+++ b/archinfo/__init__.py
@@ -21,7 +21,7 @@ from .arch_aarch64 import ArchAArch64
 from .arch_amd64 import ArchAMD64
 from .arch_arm import ArchARM, ArchARMCortexM, ArchARMEL, ArchARMHF
 from .arch_mips32 import ArchMIPS32
-from .arch_mips64 import ArchMIPS64
+from .arch_mips64 import ArchMIPS64, ArchMIPSN32
 from .arch_pcode import ArchPcode
 from .arch_ppc32 import ArchPPC32
 from .arch_ppc64 import ArchPPC64
@@ -43,6 +43,7 @@ __all__ = [
     "ArchError",
     "ArchMIPS32",
     "ArchMIPS64",
+    "ArchMIPSN32",
     "ArchNotFound",
     "ArchPPC32",
     "ArchPPC64",
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -863,11 +863,12 @@ class ArchNotFound(Exception):
     pass
 
 
-def arch_from_id(ident: str, endness=Endness.ANY, bits="") -> Arch:
+def arch_from_id(ident: str, endness=Endness.ANY, bits: str | int = "") -> Arch:
     """
     Take our best guess at the arch referred to by the given identifier, and return an instance of its class.
 
-    You may optionally provide the ``endness`` and ``bits`` parameters (strings) to help this function out.
+    You may optionally provide the ``endness`` and ``bits`` parameters to help this function out. ``bits`` is
+    either a number of bits or a string containing one, which is what an ELF class is.
     """
     if bits == 64 or (isinstance(bits, str) and "64" in bits):
         bits = 64
--- a/archinfo/arch_mips64.py
+++ b/archinfo/arch_mips64.py
@@ -198,24 +198,12 @@ class ArchMIPS64(Arch):
 
 class ArchMIPSN32(ArchMIPS64):
     """
-    The MIPS n32 ABI, and the older O64 ABI alongside it.
+    The MIPS n32 ABI and the older O64 ABI.
 
-    Both put a 64-bit instruction stream in an ELFCLASS32 container: the registers and the
-    instructions are those of ArchMIPS64, while pointers, GOT slots, relocation entries and
-    symbol table entries stay 32-bit. Every ELF structure an n32 object carries is an Elf32_*
-    one, so ``bits`` -- which is what decides how wide a word CLE reads and writes -- is 32,
-    and only the decoding side is inherited from the 64-bit architecture.
+    Both MIPS n32 and O64 use 64-bit instructions. The registers and the instructions are the same as in ArchMIPS64,
+    while pointers, GOT slots, relocation entries and symbol table entries are 32-bit.
 
-    Picking ArchMIPS64 for these instead would decode correctly and then corrupt the object:
-    CLE would stride the GOT by 8 rather than 4, read an 8-byte implicit addend out of a
-    4-byte REL slot, and write 8 bytes back over the neighbouring word.
-
-    ``bits`` is therefore the width of a pointer here and nothing else. It is not the width of a
-    register, and a consumer that needs one -- to slot an argument into ``a0``, say -- has to read
-    the size out of the register file (``registers[name]`` and ``Register.size``) or state it
-    itself. n32 is the architecture that separates the two, so it is the one that catches a
-    consumer conflating them, and on a little-endian target it will not: the two disagree only
-    about which half of a 64-bit register a 32-bit value occupies.
+    Every ELF structure an n32 object carries is an Elf32_*, so ArchMIPSN32.bits is 32.
     """
 
     def __init__(self, endness=Endness.BE):
--- a/archinfo/arch_mips64.py
+++ b/archinfo/arch_mips64.py
@@ -209,6 +209,13 @@ class ArchMIPSN32(ArchMIPS64):
     Picking ArchMIPS64 for these instead would decode correctly and then corrupt the object:
     CLE would stride the GOT by 8 rather than 4, read an 8-byte implicit addend out of a
     4-byte REL slot, and write 8 bytes back over the neighbouring word.
+
+    ``bits`` is therefore the width of a pointer here and nothing else. It is not the width of a
+    register, and a consumer that needs one -- to slot an argument into ``a0``, say -- has to read
+    the size out of the register file (``registers[name]`` and ``Register.size``) or state it
+    itself. n32 is the architecture that separates the two, so it is the one that catches a
+    consumer conflating them, and on a little-endian target it will not: the two disagree only
+    about which half of a 64-bit register a 32-bit value occupies.
     """
 
     def __init__(self, endness=Endness.BE):
--- a/archinfo/arch_mips64.py
+++ b/archinfo/arch_mips64.py
@@ -221,10 +221,12 @@ class ArchMIPSN32(ArchMIPS64):
     def __init__(self, endness=Endness.BE):
         super().__init__(endness)
         if endness == Endness.BE:
+            self.pcode_id = "MIPS:BE:64:64-32addr"
             self.triplet = "mips64-linux-gnuabin32"
             self.linux_name = "mipsn32"
             self.ida_name = "mips64b"
         else:
+            self.pcode_id = "MIPS:LE:64:64-32addr"
             self.triplet = "mips64el-linux-gnuabin32"
             self.linux_name = "mipsn32el"
 
--- a/archinfo/arch_mips64.py
+++ b/archinfo/arch_mips64.py
@@ -196,5 +196,41 @@ class ArchMIPS64(Arch):
     elf_tls = TLSArchInfo(1, 16, [], [0], [], 0x7000, 0x8000)
 
 
+class ArchMIPSN32(ArchMIPS64):
+    """
+    The MIPS n32 ABI, and the older O64 ABI alongside it.
+
+    Both put a 64-bit instruction stream in an ELFCLASS32 container: the registers and the
+    instructions are those of ArchMIPS64, while pointers, GOT slots, relocation entries and
+    symbol table entries stay 32-bit. Every ELF structure an n32 object carries is an Elf32_*
+    one, so ``bits`` -- which is what decides how wide a word CLE reads and writes -- is 32,
+    and only the decoding side is inherited from the 64-bit architecture.
+
+    Picking ArchMIPS64 for these instead would decode correctly and then corrupt the object:
+    CLE would stride the GOT by 8 rather than 4, read an 8-byte implicit addend out of a
+    4-byte REL slot, and write 8 bytes back over the neighbouring word.
+    """
+
+    def __init__(self, endness=Endness.BE):
+        super().__init__(endness)
+        if endness == Endness.BE:
+            self.triplet = "mips64-linux-gnuabin32"
+            self.linux_name = "mipsn32"
+            self.ida_name = "mips64b"
+        else:
+            self.triplet = "mips64el-linux-gnuabin32"
+            self.linux_name = "mipsn32el"
+
+    bits = 32
+    name = "MIPSN32"
+    qemu_name = "mipsn32el"
+    linux_name = "mipsn32el"
+    triplet = "mips64el-linux-gnuabin32"
+    # n32 keeps the 64-bit registers but narrows the C long and the pointer.
+    sizeof = {"short": 16, "int": 32, "long": 32, "long long": 64}
+    # Two 32-bit words of thread pointer head, as on any 32-bit MIPS.
+    elf_tls = TLSArchInfo(1, 8, [], [0], [], 0x7000, 0x8000)
+
+
 register_arch([r".*mipsel.*|.*mips64el|.*mipsel64"], 64, Endness.LE, ArchMIPS64)
 register_arch([r".*mips64.*|.*mips.*"], 64, Endness.ANY, ArchMIPS64)
--- a/tests/test_mips.py
+++ b/tests/test_mips.py
@@ -35,6 +35,18 @@ class TestArchMIPSN32(unittest.TestCase):
         assert ArchMIPSN32(Endness.LE).memory_endness == Endness.LE
         assert ArchMIPSN32(Endness.LE).struct_fmt() == "<I"
 
+    def test_argument_registers_stay_64_bit(self):
+        # bits == 32 is the pointer width. Anything that slots an argument into one of these
+        # registers has to take the width from the register file, not from bits: a consumer that
+        # divides bits by eight describes a0 as four bytes and, on big-endian, then writes a 32-bit
+        # argument into the sign-extension half that the callee never reads.
+        for endness in (Endness.BE, Endness.LE):
+            arch = ArchMIPSN32(endness)
+            assert arch.bytes == 4
+            for name in ["a0", "a1", "a2", "a3", "a4", "a5", "a6", "a7", "v0", "v1"]:
+                assert arch.registers[name][1] == 8, name
+                assert arch.registers[name] == ArchMIPS64(endness).registers[name], name
+
     def test_c_types(self):
         arch = ArchMIPSN32(Endness.BE)
         assert arch.sizeof["long"] == 32
--- a/tests/test_mips.py
+++ b/tests/test_mips.py
@@ -1,0 +1,45 @@
+# pylint:disable=no-self-use
+from __future__ import annotations
+
+import unittest
+
+from archinfo import ArchMIPS64, ArchMIPSN32
+from archinfo.arch import Endness
+
+
+class TestArchMIPSN32(unittest.TestCase):
+    """
+    n32 and O64 are 64-bit MIPS instruction streams with 32-bit pointers, so ArchMIPSN32 has to
+    disagree with ArchMIPS64 about the word size and agree with it about everything else.
+    """
+
+    def test_word_size_is_32_bit(self):
+        arch = ArchMIPSN32(Endness.BE)
+        assert arch.bits == 32
+        assert arch.bytes == 4
+        # This is what decides how wide a word CLE reads out of and writes back into a
+        # relocation slot; n32 relocation entries, GOT slots and pointers are all 4 bytes.
+        assert arch.struct_fmt() == ">I"
+
+    def test_instruction_set_is_64_bit(self):
+        arch = ArchMIPSN32(Endness.BE)
+        assert arch.vex_arch == "VexArchMIPS64"
+        assert arch.name != ArchMIPS64.name
+        # The hardware registers stay 64-bit even though a pointer does not.
+        assert arch.registers["sp"][1] == 8
+        assert arch.registers["ra"][1] == 8
+        assert arch.registers["pc"][1] == 8
+
+    def test_endness(self):
+        assert ArchMIPSN32(Endness.BE).memory_endness == Endness.BE
+        assert ArchMIPSN32(Endness.LE).memory_endness == Endness.LE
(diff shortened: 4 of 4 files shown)


zardus:

Validation record for head `39bab86b70890c41bd57a6fff55e467c73986007` against merge base `bf85c7e47bb469878c564480e28c677abbe35acc`.

- Regression: `python -m pytest tests/test_mips.py` — 7 passed on the head; the same file at the merge base cannot be collected at all, `ImportError: cannot import name 'ArchMIPSN32' from 'archinfo'`. The `pcode_id` test added at this head was also run against the previous head `9e9d9e26498bb5553ff5450a6764be161138f51d`, where it fails with `AssertionError: assert 'MIPS:BE:64:default' == 'MIPS:BE:64:64-32addr'`
- Full suite: `python -m pytest tests` — 26 passed, 0 skipped on the head, in a virtualenv carrying pyvex and pypcode so nothing skipped
- Thin environment: archinfo declares no dependencies, so its own CI job has neither. With `pypcode` and `pyvex` both refused by a `find_spec` meta-path hook — proved to raise `ImportError` before the run — the suite is 15 passed, 11 skipped, no error
- Hooks and CI checks: every configured pre-commit hook over all files, twice, leaving the tree unchanged; `run-ci-diff-checks.py` reports no lint or type regression against the merge base, with `arch_mips64.py` at pylint 9.84 → 9.88 and pyright badness 0.150 → 0.122
- Environment: Python 3.12.14, pytest 9.1.1, pyvex 9.3.3 and pypcode 4.0.0 from PyPI, in a virtualenv created for this worktree. `archinfo.__file__` was asserted to resolve inside the worktree before every run

Caveats: this is archinfo's own suite only — no workspace-wide gate, and no consumer repository was rebuilt or retested. `ArchMIPSN32` is deliberately not registered for name-based lookup, so nothing resolves it by name and only an explicit construction or a loader that selects it reaches the class. The n32 and O64 fixtures are in angr/binaries#206; nothing in this suite loads a file.

### Merge order, and what an incomplete chain does

The four pull requests in this chain are `angr/archinfo#375` (defines `ArchMIPSN32`),
`angr/pyvex#576` (teaches the lifter that name), `angr/cle#795` (resolves the ELF variant)
and `angr/angr#6982` (calling conventions, plus a `CFGFast` assertion). **They must land
together, in that order: archinfo#375, then pyvex#576, then cle#795, then angr#6982.**

Merging the loader and the architecture definition ahead of the lifter is the dangerous
partial state, and it is also the natural reading order. In it `ArchMIPSN32` resolves, the
name reaches a lifter with no entry for it, every block comes back `Ijk_NoDecode`, and angr
recovers **0 blocks and 0 functions** on every object in this population — worse than the
unfixed baseline, which at least recovers the 32-bit-decodable parts.

The figures below were measured at archinfo `9e9d9e26498bb5553ff5450a6764be161138f51d`. The
only commit added since, `39bab86b70890c41bd57a6fff55e467c73986007`, points `ArchMIPSN32.pcode_id`
at `MIPS:{BE,LE}:64:64-32addr` instead of inheriting `MIPS:{BE,LE}:64:default` from `ArchMIPS64`,
and adds two tests. Engine selection is `isinstance(project.arch, archinfo.ArchPcode)`
(`angr/factory.py:55`), which `ArchMIPSN32` is not — it keeps `vex_arch = "VexArchMIPS64"` — and a
full-text search of angr, cle, pyvex and angr-management finds no reader of `pcode_id` and no
caller of `Arch.pcode_arch()`; the identifier occurs in archinfo alone. So these figures stand
unchanged at the new head, whose only observable effect is that an explicit
`ArchMIPSN32(...).pcode_arch()` now returns an `ArchPcode` with `bits == 32` rather than 64.

Measured 2026-08-28, at the heads named in the table, one process per object:

| arm | archinfo | pyvex | cle | angr |
| --- | --- | --- | --- | --- |
| master | `bf85c7e47bb469878c564480e28c677abbe35acc` | `bdd5441035e02920eaa72c1c3cf9a4f0d572104d` | `d2ecea068794d20b1f14d90eecc1bc4bc4cfa431` | `0e18fa25ef83c3db9813fe5a156da0013052e0e5` |
| loader + arch only | #375 `9e9d9e26498bb5553ff5450a6764be161138f51d` | master | #795 `4aaf40ece48f07e863507e80769984ae138d11c2` | master |
| all four | #375 `9e9d9e26…` | #576 `2bf2cff41aa1beb9a502a311a5154411f8d41158` | #795 `4aaf40ec…` | #6982 `3ac333750ad5b2ce76a62e6f3818c4000dbcf1d8` |

Population: 16 ELF objects, every one `ELFCLASS32` `EM_MIPS`. Fourteen declare a MIPS III
ISA in `e_flags` — ten big-endian O64 (`0x20002001`) and four big-endian n32
(`0x20000027`) — and two are ordinary little-endian o32 (`0x1001`), held as controls. The
corpus is not public, so they are cited by digest below. Metric: DWARF line-table addresses
covered by no recovered block, under `CFGFast(normalize=True, resolve_indirect_jumps=True)`.

Over the fourteen MIPS III objects, of 261,039 line addresses:

| arm | missed | blocks | functions | missed function starts |
| --- | ---: | ---: | ---: | ---: |
| master | 59,268 | 151,093 | 41,104 | 684 |
| loader + arch only | 261,039 | 0 | 0 | 9,131 |
| all four | 2,366 | 167,600 | 15,989 | 16 |

- The partial state misses **100%** of every object's own line table, on all fourteen. The
  two o32 controls are untouched by it, so the damage is confined to the objects the chain
  is about.
- The complete chain removes **96.0%** of the master survivors, with block counts up on
  every one of the fourteen and no object regressing. Example: `a9271ebdaaef5111…` goes
  6,969 → 219 missed and 14,851 → 16,628 blocks; `761f33eb0e35fca5…` goes 46 → 1 missed and
  360 → 373 blocks.
- The function count falls because master's is inflated: each `Ijk_NoDecode` boundary starts
  a fresh pseudo-function. Two independent counters move the right way — declared function
  entry points with no covering block go 684 → 16, and FDE starts 60 → 0.
- Controls: the two o32 objects report identical block counts, function counts and
  missed-address **sets** under every arm measured here, so nothing in the chain reaches
  ordinary MIPS32.
- Determinism: the three arms were run in both orders (master → partial → chain, then chain
  → partial → master). All sixteen objects agree on every field in both passes.

**`angr/pyvex#567` overlaps this chain; do not add the two figures.** Measured alone against
the same master (pyvex `dca28718d29976fa0a775199076c8197bd177f89`, vex
`561795c17f97a008108d8642d4854c56e79513a2`, `libpyvex.so` rebuilt for it), it takes the
fourteen from 59,268 to 47,526 — 11,742 addresses, **19.8%**, per object 13.8%–32.6%. Added
on top of the complete four-PR chain it changes **nothing at all**: 2,366 both ways, object
for object, byte for byte. It repairs the discarded prefix, which the chain already avoids
by resolving the right architecture. Its two o32 controls are identical under every arm, so
it has no collateral effect on ordinary MIPS32 either.

Method notes: each arm asserts the resolved `__file__` of `angr`, `cle`, `pyvex` and
`archinfo` lies inside the intended worktree before measuring, because a `PYTHONPATH` entry
that does not exist is skipped silently and falls through to the editable install. The
pyvex#567 native build was made in its own worktree and gated on a behavioural probe
(`ld $a0, -0x3098($v1)` at `0x80010938` lifts to `size=8` under #567 and `size=0` under
master), not on a resolved path; the shared checkout's prebuilt `libpyvex.so` was
hash-checked before and after that build and is unchanged.

Objects, `sha256`, all ELF:

```
O64  a9ab23096d37b67368d0ff00e6c37059e631f8f04c5b337a98176ac4f70ff377
O64  a9271ebdaaef51110e0a6fc23b1f3949e447ec825ae47fb4c9dc2b9c0d1b2634
O64  cbd909ae5fe39e44506a4975d106e044289706afdd659d37463a247211261dbf
O64  3b08741fdbc63589870323fe5f40373d543f34fdacb487218aad3ffcb65381dc
O64  7e65689497320f65d5951375c27116f79e5cec6cb20970169e5dc6028c242a78
O64  b860cfbe35c32968c8194bb29d223076964b452cd38ebb16e0bcf3f00328af19
O64  f2afd5b17235d3c9a10c993c03fad3d6883d89d2b861022f317d270e181f4c5e
O64  ebea75270e2c9d00170704b675d2b18c45b1b2aebbe34cdb13703bcec54b2f4c
O64  b634a05e39a6000fa52d276ba5518ffb67b03b3ddc31f79b4378bb319a41c89b
O64  a7d9003ec00bd8bb499573617fb51b3ce5b2d9a18c746f404606cce203c182e8
n32  acbc7e75858097e906084988261e7cd646cb4ee507cd7573d8719617bd45b196
n32  d254234ebbc95e26bba689de241193439828e975bd0945170d4df664c75588df
n32  ca2055345f9d85b4c66209e67a59facbc68cb5820c230c6f154123c47eef5b76
n32  761f33eb0e35fca5d108e729bd8066909163b3212b6238834a3cb922224b6739
o32  728368792e9a34a3a7a78ed9f0ccdfaee3f96ffd15ef09c8e6678ca6d098e882
o32  04b8ade9cac4172a71d90fdedf7bc7545fa0e32bd69b7d0c77fd8f835f8e65f6
```

**Public reproducer for the ordering**, on the fixtures in `angr/binaries#206` at
`f54a80068c7e67e1a48c44eab772a03ec1c74476`, no corpus needed. For each of
`tests/mipsn32/{n32_be_static,o64_el_static,n32_el_dynamic}`, run
`angr.Project(path, auto_load_libs=False).analyses.CFGFast(normalize=True)` and count
`cfg.model.nodes()` and `cfg.kb.functions`:

| arm | n32_be_static | o64_el_static | n32_el_dynamic |
| --- | --- | --- | --- |
| master | `MIPS32`, 13 blocks / 8 functions | `MIPS32`, 13 / 8 | `MIPS32`, 35 / 20 |
| archinfo#375 + cle#795, pyvex at master | `MIPSN32`, **0 / 0** | `MIPSN32`, **0 / 0** | `MIPSN32`, **0 / 0** |
| all four | `MIPSN32`, 13 / 6 | `MIPSN32`, 13 / 6 | `MIPSN32`, 35 / 15 |

twizmwazin:
Can ArchPCode express that?

zardus:

Partly — Ghidra models n32 as a first-class language and `ArchPcode` picks up the right `bits`, but the 64/32 split it would be expressing is only the ldefs attribute, pypcode discards the part that makes it real, and no n32 ELF can reach an `ArchPcode` today.

**The language exists.** [`mips.ldefs:176-199`](https://github.com/angr/pypcode/blob/master/pypcode/processors/MIPS/data/languages/mips.ldefs#L176-L199) defines `MIPS:BE:64:64-32addr` with `size="32"`, `<truncate_space space="ram" size="4"/>`, a dedicated `n32` compiler spec (`mips64_32_n32.cspec`) and `<external_name tool="qemu" name="qemu-mipsn32"/>`. `MIPS.opinion:114-116` even constrains on `EF_MIPS_ABI2`. LE, micro and R6 variants exist too, six ids in all with `size="32"`.

**And `ArchPcode` reads it correctly.** `self.bits = int(language.size)` (`arch_pcode.py:38`), `self.bytes = self.bits // self.byte_width` (`arch.py:181`), register file from `pypcode.Context(language).registers`. So:

```
>>> a = archinfo.ArchPcode("MIPS:BE:64:64-32addr")
>>> a.bits, a.bytes, len(a.register_list), a.registers["sp"]
(32, 4, 521, (232, 8))
```

32-bit pointers over the 64-bit register file — structurally the same thing `ArchMIPSN32` hand-writes.

**But the semantics are not 32-bit.** `MIPS:BE:64:64-32addr` and `MIPS:BE:64:default` name the *same* `mips64be.sla` (identical file, sha256 `47cc7e1bc4113459…`); `mips64be.slaspec` never defines `MIPS64_32ADDRS`, so `mips.sinc:179` is `define space ram type=ram_space size=8`. The narrowing lives only in `<truncate_space>`, which is applied by `AddrSpaceManager::truncateSpace` — and pypcode's `Context` ctor (`pypcode/__init__.py:241`, `pypcode_native.cpp:253-270`) never calls it, while the `AddrSpace` binding exposes only `name`, no `getAddrSize()`. Translating `df a8 00 00` under both ids gives byte-identical p-code with an 8-byte address varnode into `LOAD`. So the answer to "64-bit instructions with 32-bit pointers" is: SLEIGH can say it, pypcode currently cannot hear it.

**And it is unreachable regardless.** `arch_from_id("em_mips", "be", 32)` returns `ArchMIPS32` (`arch_mips32.py:287`), so cle's p-code fallback at [`elf.py:350`](https://github.com/angr/cle/blob/master/cle/backends/elf/elf.py#L350) — which only fires on `ArchNotFound` — never runs. `arch_pcode.py` calls `register_arch` nowhere, so `arch_from_id` can never return an `ArchPcode` at all. The `e_flags` branch in angr/cle#795 is required either way; only its return value would differ.

**What it would cost.** `ArchPcode` has `vex_arch = None` and `cs_arch = None`, so every n32 object moves off libVEX onto `UberEnginePcode` (`angr/factory.py:55`, warned as experimental, with open `NotImplementedError`s in its lifter). `DEFAULT_CC`/`CC`/`SYSCALL_CC` are name-keyed and `unify_arch_name` cannot map a MIPS Sleigh id — `ARCH_NAME_ALIASES["MIPS64"]` is `[]` (`calling_conventions.py:2708`) — so the CC would go into the hand-written `manual_cc_mapping` in `angr/engines/pcode/cc.py:120`, which has no syscall-table equivalent; `SimLinux` falls to `syscall_abis = []`; and the MIPS GP resolver is `isinstance`-gated on `(ArchMIPS32, ArchMIPS64)` (`mips_elf_fast.py:59`), which the subclass satisfies and an `ArchPcode` does not.

**To make it the right answer** you'd need pypcode to apply `truncate_space` and expose the address-space size, `ArchPcode` to read that rather than only `size=`, and a registration path from `e_flags` to a language id. That is worth doing on its own — `ArchPcode` gets `initial_sp = 0x7fffffffffffffff` on a `bytes=4` arch today (`arch_pcode.py:120-123` sizes it from the `sp` register), `instruction_alignment` is hardcoded to 1 against SLEIGH's `alignment=2`, and `branch_delay_slot` is only set for 32-bit sparc — but it is a pypcode/`ArchPcode` change, not this one.

One thing your question did surface in this patch: `ArchMIPSN32` inherits `pcode_id = "MIPS:{BE,LE}:64:default"` from `ArchMIPS64`, when `MIPS:{BE,LE}:64:64-32addr` is exactly right for it. Happy to correct that here.

zardus:

Corrected, in `39bab86b70890c41bd57a6fff55e467c73986007`. `ArchMIPSN32` now names `MIPS:BE:64:64-32addr` / `MIPS:LE:64:64-32addr` rather than inheriting `MIPS:{BE,LE}:64:default` from `ArchMIPS64`. `tests/test_mips.py` pins both ids so the inherited value cannot come back silently, and a second test — skipped when pypcode is absent, which archinfo's own job is — resolves them through `pcode_arch()`, which is what would catch a typo in the string.

Nothing downstream moves. Engine selection is `isinstance(project.arch, archinfo.ArchPcode)` in angr's `factory.py:55`; `ArchMIPSN32` is not one and keeps `vex_arch = "VexArchMIPS64"`. A full-text search of angr, cle, pyvex and angr-management finds no reader of `pcode_id` and no caller of `Arch.pcode_arch()` — the identifier occurs in archinfo alone. `register_pcode_arch_default_cc` keys on `arch.name` behind the same `isinstance` guard, and cle's `_extract_pcode_arch` builds an `ArchPcode` from a pypcode language object rather than reading an id off an existing arch. The one observable difference is that an explicit `ArchMIPSN32(...).pcode_arch()` now returns `bits == 32`, `bytes == 4` instead of 64 and 8, with the `n32` compiler spec available on the language.

The corpus and fixture figures therefore stand at the new head; the validation record is re-keyed to it in place, with the local results and the argument for the carry-over.

Merge order is unchanged: archinfo 375, then pyvex 576, then cle 795, then angr 6982. Merging the loader and the architecture definition ahead of the lifter still drops recovery to zero blocks on MIPS-III objects.

zardus:

Reading `tests/mipsn32/n32_el_dynamic` — a 64-bit MIPS instruction stream in an ELFCLASS32 container — with each MIPS architecture archinfo offers, before and after this change.

**Before** — `ArchMIPS32` cannot decode the 64-bit store, and `ArchMIPS64` reads the 4-byte GOT slots eight at a time:

<details><summary>archinfo master</summary>

```
archinfo: archinfo at master f92307b
n32_el_dynamic: ELFCLASS32 EM_MIPS e_flags=0x80000027 (EF_MIPS_ABI2, the n32 mark)
.got at 0x10a30: sh_entsize=4, 68 bytes = 17 slots
instruction at 0x6c4: 0800bcff

architecture   bits bytes vex_arch         struct_fmt  0x6c4 decodes to
ArchMIPS32       32     4 VexArchMIPS32    '<I'        <decodes to nothing>
ArchMIPS64       64     8 VexArchMIPS64    '<Q'        sd $gp, 8($sp)
ArchMIPSN32    AttributeError: module 'archinfo' has no attribute 'ArchMIPSN32'

the first four .got slots, read at each architecture's word width:
    ArchMIPS32   stride 4: 0x00000000 0x80000000 0x000006c0 0x00010a10
    ArchMIPS64   stride 8: 0x8000000000000000 0x00010a10000006c0 0x000008b0000109f0 0x00010a2400010000
    ArchMIPSN32  unavailable
```

</details>

**After** — `ArchMIPSN32` decodes as MIPS64 and strides the GOT as MIPS32:

```
archinfo: archinfo at this branch
n32_el_dynamic: ELFCLASS32 EM_MIPS e_flags=0x80000027 (EF_MIPS_ABI2, the n32 mark)
.got at 0x10a30: sh_entsize=4, 68 bytes = 17 slots
instruction at 0x6c4: 0800bcff

architecture   bits bytes vex_arch         struct_fmt  0x6c4 decodes to
ArchMIPS32       32     4 VexArchMIPS32    '<I'        <decodes to nothing>
ArchMIPS64       64     8 VexArchMIPS64    '<Q'        sd $gp, 8($sp)
ArchMIPSN32      32     4 VexArchMIPS64    '<I'        sd $gp, 8($sp)

the first four .got slots, read at each architecture's word width:
    ArchMIPS32   stride 4: 0x00000000 0x80000000 0x000006c0 0x00010a10
    ArchMIPS64   stride 8: 0x8000000000000000 0x00010a10000006c0 0x000008b0000109f0 0x00010a2400010000
    ArchMIPSN32  stride 4: 0x00000000 0x80000000 0x000006c0 0x00010a10
```

</details>

zardus:

Heads-up from the [angr/mono#3](https://github.com/angr/mono/pull/3) rollup: this pull request was **skipped** during assembly, and the four pull requests stacked on it were not.

#365 and #375 both rewrite the signature and the opening docstring paragraph of `arch_from_id` in `archinfo/arch.py`. They merge in number order, so #365 lands and #375 is dropped with `conflicts with the applied set: archinfo/arch.py`. angr/binaries#206, angr/pyvex#576, angr/cle#795 and angr/angr#6982 all applied on top of an archinfo that therefore has no `ArchMIPSN32`, and `import angr` fails at `calling_conventions.py:2602`:

```
AttributeError: module 'archinfo' has no attribute 'ArchMIPSN32'. Did you mean: 'ArchMIPS32'?
```

which takes out all five `pyinstaller` jobs, the whole `angr-management` suite, and 5 of 7 tests in cle's `tests/test_mips_n32.py`.

Nothing here is wrong with the change itself — merged at head `39bab86b` on base `b8ffe857` together with #364 and #376 it applies clean, archinfo's own 35 tests pass, and cle's `test_mips_n32.py` goes to 7 passes. #365 is the side being held out of the rollup selection meanwhile. A rebase of either pull request past the other resolves it and the exclusion is retired; the conflict is a signature-and-docstring overlap, not a disagreement about behaviour.

session: sharpen

ltfish:
Dear agent, your docstring for the `ArchMIPSN32` class is incredibly verbose and redundant. Please try to write it in plain and human-understandable English.

ltfish:
Adding this new arch def looks good to me.

ltfish:
I'm fixing it.

PR #377, opened 2026-08-30
Populate Arch.registers for p-code architectures without pyvex

zardus:

### Problem

`pip install archinfo[pcode]` installs pypcode and not pyvex, because archinfo
declares no runtime dependencies. In that configuration every `ArchPcode` comes
back with an empty register file, and two of archinfo's own tests in
`tests/test_pcode.py` fail with `KeyError: 'ip'` and `KeyError: 'sp'`:

```
>>> archinfo.ArchPcode("x86:LE:64:default").registers
{}
```

`register_names` is empty too, `artificial_registers` is never set at all, and
this is all 187 languages pypcode ships.

### Root cause

`Arch.__init__` gates the whole block on `if self.register_list and _pyvex is not
None`. Only one thing inside it needs pyvex: `vex_ffi.guest_offsets`, which fills
in the offset of a register that does not have one, and only a VEX architecture
has any. An `ArchPcode` register already carries the offset pypcode gave it.

### Fix

Do the `guest_offsets` lookup when there is a VEX architecture to look it up for,
and build the collections whenever the offsets are already known. A VEX
architecture with no pyvex still gets nothing, as before, and nothing changes at
all when pyvex is installed.

### Testing

`tests/test_pcode.py::TestArchPcode::test_registers_without_pyvex` patches
`archinfo.arch._pyvex` to `None`, which is what the failed import leaves behind,
so it runs wherever the suite runs; it fails on master and passes here. The two
tests that were already failing without pyvex pass. Across all 187 p-code
languages `ArchPcode` now produces
exactly the register state it produces with pyvex installed, and no VEX
architecture changes in either configuration.

Validation: https://github.com/angr/archinfo/pull/377#issuecomment-5469107487

session: sharpen

### DIFF
--- a/archinfo/arch.py
+++ b/archinfo/arch.py
@@ -203,11 +203,11 @@ class Arch:
                 self.ret_instruction = reverse_ends(self.ret_instruction)
                 self.nop_instruction = reverse_ends(self.nop_instruction)
 
-        if self.register_list and _pyvex is not None:
-            (_, _), max_offset = max(_pyvex.vex_ffi.guest_offsets.items(), key=lambda x: x[1])
-            max_offset += self.bits
+        if self.register_list and (_pyvex is not None or not self.vex_support):
             # Register collections
-            if isinstance(self.vex_arch, str):
+            if _pyvex is not None and isinstance(self.vex_arch, str):
+                (_, _), max_offset = max(_pyvex.vex_ffi.guest_offsets.items(), key=lambda x: x[1])
+                max_offset += self.bits
                 va = self.vex_arch[7:].lower()  # pylint: disable=unsubscriptable-object
                 for r in self.register_list:
                     if r.vex_offset is None:
--- a/tests/test_pcode.py
+++ b/tests/test_pcode.py
@@ -1,6 +1,7 @@
 # pylint:disable=missing-class-docstring,no-self-use
 import pickle
 import unittest
+from unittest.mock import patch
 
 from archinfo import ArchError, ArchPcode, ArchS390X, Endness, arch_from_id
 
@@ -42,6 +43,13 @@ class TestArchPcode(unittest.TestCase):
         assert ArchPcode("z80:LE:16:default").initial_sp == 0x7FFF
         assert ArchPcode("68000:BE:32:default").initial_sp == 0x7FFFFFFF
         assert ArchPcode("x86:LE:64:default").initial_sp == 0x7FFFFFFFFFFFFFFF
+
+    def test_registers_without_pyvex(self):
+        """A p-code register file comes from pypcode, so archinfo still builds it with no pyvex installed."""
+        with patch("archinfo.arch._pyvex", None):
+            arch = ArchPcode("x86:LE:64:default")
+        assert arch.registers["ip"] == arch.registers["rip"]
+        assert arch.registers["sp"] == arch.registers["rsp"]
 
     def test_arch_bad_langid(self):
         with self.assertRaises(ArchError):


zardus:

Validation record for head `e9a5e026c28c5d4433ffc7fbf2a440895467800b` against
baseline `7e72c6caed422219d97af234f7acec96b37535c4`.

Two configurations, both Python 3.12.13, in throwaway virtualenvs with archinfo
installed editable from the branch worktree, running `python -m pytest tests/ -q`:

| | pypcode 4.0.0 and pyvex 9.3.3 | pypcode 4.0.0, no pyvex |
| --- | --- | --- |
| baseline | 31 passed | 2 failed, 27 passed, 2 skipped |
| baseline with this branch's tests | 1 failed, 31 passed | 3 failed, 27 passed, 2 skipped |
| head | 32 passed | 30 passed, 2 skipped |

34 subtests pass in every cell. The baseline failures without pyvex are
`tests/test_pcode.py::TestArchPcode::test_arch_with_program_counter` and
`::test_arch_with_narrow_stack_pointer`; the new
`::test_registers_without_pyvex` fails at baseline in both configurations, which
is what puts it in reach of the existing CI job. The two skips are the
pre-existing `requires_pyvex` cases in `tests/test_mips.py`.

Equivalence sweep over the 22 registered architectures and all 187 pypcode
languages, dumping `registers`, `register_names`, `register_size_names`,
`subregister_map`, `default_register_values`, `entry_register_values`,
`default_symbolic_registers`, `argument_registers`, `persistent_regs`,
`concretize_unique_registers`, `artificial_registers`,
`artificial_registers_offsets`, `ip_offset`, `sp_offset`, `bp_offset`,
`lr_offset`, `ret_offset` and `initial_sp` for each, under `PYTHONHASHSEED=0`:

| comparison | VEX arches | p-code languages |
| --- | --- | --- |
| with pyvex, baseline vs head | 0 of 22 differ | 0 of 187 differ |
| without pyvex, baseline vs head | 0 of 22 differ | 187 of 187 differ |
| without pyvex vs with pyvex, at baseline | 21 of 22 differ | 187 of 187 differ |
| without pyvex vs with pyvex, at head | 21 of 22 differ | 0 of 187 differ |

So the pyvex-installed configuration is untouched, no VEX architecture moves in
either configuration, and every p-code language now agrees with the
pyvex-installed configuration. The 21 of 22 rows are the unchanged behaviour of a
VEX architecture with no pyvex to supply its register offsets, at both revisions;
the 22nd is `ArchSoot`, which has no register list. `ArchMIPSN32` is exported but
not registered, so it is not among the 22; it is a VEX architecture and the 0 of
22 row covers its case. `PYTHONHASHSEED=0` is needed because
`artificial_registers_offsets` is built by iterating a set, so its order varies
between processes on master as well as here.

Of those 18 fields, 13 are byte-identical in the pypcode-only configuration at
both revisions, on all 187 languages: `argument_registers`, `sp_offset`,
`initial_sp` and the rest are already right without pyvex. Two go from empty to
populated, `registers` and `register_names`. Two go from absent or `None` to an
empty container: `artificial_registers`, which raised `AttributeError` before,
and `artificial_registers_offsets`. One changes value -- `bp_offset` moves on 6
x86 p-code languages, `x86:LE:32:default` 16 to 20, `x86:LE:64:default` 32 to 40
and four siblings, because `ArchPcode` sets `bp_offset` from the stack pointer
and the block now overrides it with the BP register's own offset. All 6 land on
the value the pyvex-installed configuration already produced.

Lint and type over `archinfo/arch.py` and `tests/test_pcode.py`, at the versions
the pre-commit configuration and the CI image pin. `ruff 0.16.4` clean; `black
26.5.1` reports both files unchanged; `pylint` with `angr/ci-settings`' `pylintrc`
rates `arch.py` 9.70/10 and `test_pcode.py` 10.00/10, each identical at both
revisions, which is what `lint.py` compares. `pyright 1.1.411` reports 19
diagnostics at both revisions with pyvex and pypcode resolved, which is CI's
environment, and 18 at both with only pypcode resolved; nothing is added in
either, `tests/test_pcode.py` contributes none, and the only movement is a
pre-existing `reportAttributeAccessIssue` on the `guest_offsets` line shifting
from 207 to 209. The complete pre-commit hook set, 24 hooks, passes over all
files.

Not run: the workspace-wide gate. Five corpus sweeps hold the shared virtualenv,
and entering `nix develop` would reinstall the editables that toolchain is
running on, so this is the scoped equivalent `angr-validate-workspace`
prescribes -- archinfo's own suite in an isolated worktree with its own
virtualenvs. angr, angr-management, cle, claripy and pyvex were not exercised;
they all hard-depend on pyvex, where this change is a no-op both by inspection
and by the 0-of-22 and 0-of-187 rows above. The with-pyvex virtualenv uses PyPI
pyvex 9.3.3 rather than a source build, because building pyvex would overwrite
the native libraries the running sweeps have mapped.

What the new test does not cover: no CI job installs pypcode without pyvex -- the
Windows and macOS jobs run `uv run pytest`, which installs neither, and the
`angr/ci-settings` image carries both. Patching `_pyvex` to `None` inside an
already-imported archinfo catches the gate being narrowed again; it cannot catch
the gate being widened too far, which in a real pypcode-only install fails at
`import archinfo`. Closing that needs a job in the missing configuration and is
worth doing separately.

zardus:

`ArchPcode("x86:LE:64:default")` in a virtualenv with pypcode installed and pyvex
absent, before and after this change.

```python
import logging
logging.disable(logging.CRITICAL)
import archinfo
a = archinfo.ArchPcode("x86:LE:64:default")
print("len(a.registers) =", len(a.registers))
print('a.registers["rip"] =', a.registers["rip"])
print('a.registers["ip"]  =', a.registers["ip"])
```

**Before** — the register file is empty, so every lookup by name raises:

<details><summary>archinfo master, 7e72c6caed422219d97af234f7acec96b37535c4</summary>

```
len(a.registers) = 0
Traceback (most recent call last):
  File "repro.py", line 6, in <module>
    print('a.registers["rip"] =', a.registers["rip"])
                                  ~~~~~~~~~~~^^^^^^^
KeyError: 'rip'
```

</details>

**After** — the same register file the pyvex-installed configuration produces:

```
len(a.registers) = 1441
a.registers["rip"] = (648, 8)
a.registers["ip"]  = (648, 8)
```

</details>

twizmwazin:
Supporting pyvex-free archinfo isn't really a concern

